PR #20, opened 2017-10-10
Change Makefile

zhouzhuojie:
Test CI


zhouzhuojie:
Finally get the CI working

- Replace phantomjs with jsdom, so that it doesn't throw timeout error connecting localhost
- Replace scss with less, so that there's no weird libsass installation issue with alpine, also it requires python, glibc, and etc.
- Remove e2e tests, so that we don't need to handle selenium, which requires java8, and it's hard to get the dependency right.
- other package.json fix

browser/flagr-ui/src/App.vue
```diff
@@ -24,7 +24,7 @@ export default {
 }
 </script>
 
-<style lang="scss">
+<style lang="less">
```
zhouzhuojie: @lucidrains I've changed from scss to less to pass CI

PR #35, opened 2017-10-16
Combine the constraints evaluation logic

zhouzhuojie:
Combines the constraints evaluation logic.

This is to make the evaluation even faster by combining the constraints together in a segment. The other benefits are - for example - to enable cross-reference on the constraints' properties.

For example, 

image


pkg/config/config.toml
```diff
@@ -1,17 +0,0 @@
...
-[Sentry]
-Enabled = false
-DSN = ""
-
-[EvalCache]
-RefreshTimeout = "59s"
-RefreshInterval = "10s"
-
-[PProf]
-Enabled = true
```
zhouzhuojie: Going to loop in the config file into env variables

pkg/entity/constraint.go
```diff
@@ -50,20 +66,33 @@ func (c *Constraint) ToExpr() (conditions.Expr, error) {
...
+func (cs ConstraintArray) ToExpr() (conditions.Expr, error) {
+	strs := make([]string, 0, len(cs))
+	for _, c := range cs {
+		s, err := c.toExprStr()
+		if err != nil {
+			return nil, err
+		}
+		strs = append(strs, s)
+	}
+	exprStr := strings.Join(strs, " AND ")
```
zhouzhuojie: Combining all the constraints with `AND` under one segment.

PR #37, opened 2017-10-16
Add put and delete for variants

zhouzhuojie:
Put and delete for variants


pkg/handler/validate.go
```diff
@@ -47,3 +48,22 @@ func validatePutDistributions(params distribution.PutDistributionsParams) *Error
...
+	err := entity.NewFlagQuerySet(repo.GetDB()).IDEq(uint(params.FlagID)).One(f)
+	if err != nil {
+		return NewError(400, "error finding flagID %v. reason %s", params.FlagID, err)
+	}
+	f.Preload(repo.GetDB())
+
+	for _, s := range f.Segments {
+		for _, d := range s.Distributions {
+			if d.VariantID == util.SafeUint(params.VariantID) && d.Percent != uint(0) {
+				return NewError(400, "error deleting variant %v. distribution %v still has non-zero distribution %v", params.VariantID, d.ID, d.Percent)
```
zhouzhuojie: Note, we cannot delete a variant if it's rolling out to some non-zero distribution

lucidrains: oo ok, yea, i can put that check on the frontend too

zhouzhuojie: yup, actually I need to remove all the distribution that has the references to it

lucidrains: ohh, so do you want it so when they delete a variant, it deletes all referenced distributions? or just not allow that until they delete the distribution?

zhouzhuojie: going to delete the referenced distribution if percent is `0`, otherwise, return 400

lucidrains: oo ok

PR #42, opened 2017-10-18
Change variant key should also change distribution

zhouzhuojie:
Change the variant key should also sync with the distribution's variant key

### DIFF
--- a/buildscripts/demo.py
+++ b/buildscripts/demo.py
@@ -1,0 +1,38 @@
+import httplib
+import random
+import json
+import datetime
+
+conn = httplib.HTTPConnection("localhost:18000")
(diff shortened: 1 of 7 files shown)


buildscripts/demo.py
```diff
@@ -0,0 +1,38 @@
...
+
+while 1:
+    t = datetime.datetime.now()
+    conn.request("POST", "/api/v1/evaluation", random_payload(), headers)
+    res = conn.getresponse()
+    print(str((datetime.datetime.now() - t).total_seconds() * 1000) + 'ms')
+
+    data = res.read()
+    index_elastic(data)
+    print(data.decode("utf-8"))
```
zhouzhuojie: Only for demo, but good to have a client example here

PR #56, opened 2017-10-26
Add api docs button and data records toggle

zhouzhuojie:
Some UI tweaks

- Adding API and Doc header links
- Adding data records toggle - which will be used to enable kafka data logging for the flag
- Adding error messages support for all the endpoint
- Replace fetch with vue-resource for consistency


lucidrains:
Nice! Vue Resource looks good!

Issue #102, opened 2018-03-12
Monitoring / Heath Check 


marceloboeira:
Considering that this would be a big part of the infrastructure (getting lots of public requests), it would be interesting to have a **health check endpoint**, in order to monitor the status. Even more if you consider autoscaling or anything of the sort (knowing when the pod is ready to receive traffic, or if its unhealthy and shouldn't receive much load). e.g.: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/

For even more control over the infrastructure, sending metrics to **statsd** would also be really awesome, in order to integrate with something like Datadog and investigate any latency with more precision.  

If its something that you guys think its okay I might be willing to work on PR for both.

zhouzhuojie:
PRs are welcome!

1) for liveness check, if you want to work on it, you can start with the swagger definition of `/health` endpoint, and then implement it in `pkg/handler`
2) for statsd, you can take a look at the new relic middleware under `config/middleware.go`

Thanks!

PR #112, opened 2018-04-06
Push http metrics with statsd

marceloboeira:
Push metrics, such as response time, response count ..., to a statsd server.

First part of #102 

## Description
Besides the default configuration, there is a negroni middleware to push http metrics to statsd. 

## Motivation and Context
Have control metrics of individual parts of the architecture, e.g.: response times, individual processes breakdown, project health in general.  

## How Has This Been Tested?
An statsd server started locally, checked if the metrics are sent.
Deployed to production on my cluster... 

## Types of changes

## Checklist:


pkg/config/middleware.go
```diff
@@ -87,3 +94,18 @@ func (a *auth) ServeHTTP(w http.ResponseWriter, req *http.Request, next http.Han
...
 }
+
+type statsdMiddleware struct {
+	StatsdClient *statsd.StatsdClient
+}
+
+func (s *statsdMiddleware) ServeHTTP(w http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
+	defer func(start time.Time) {
+		response := w.(negroni.ResponseWriter)
+		status := strconv.Itoa(response.Status())
```
zhouzhuojie: You are aggregating all requests. Maybe you can extend the scope and add the route path to the stats. For example, getting the path from `r.URL.Path` or `r.URL.RawPath`, and add path to the statsd string.

```
path := r.URL.Path
// e.g. /api/v1/evaluation, you may want to replace '/' with something else for safe escapes.

s.StatsdClient.Incr(fmt.Sprintf("http.requests.%s.%s.count", path, status), 1)
s.StatsdClient.PrecisionTiming(fmt.Sprintf("http.requests.%s.duration", path), time.Since(start)
```

Or even better, use a statsd client that supports `tags`. It's nice to just tag the count and duration with path.

https://godoc.org/gopkg.in/alexcesaro/statsd.v2

pkg/config/config.go
```diff
@@ -8,19 +8,22 @@ import (
 	"github.com/evalphobia/logrus_sentry"
 	raven "github.com/getsentry/raven-go"
 	newrelic "github.com/newrelic/go-agent"
+	"github.com/quipo/statsd"
```
zhouzhuojie: For new dependencies, can you also add it to Gopkg.toml and check in the vendor for faster build time?

zhouzhuojie:
Thanks, this is awesome to have some statsd support! Is it compatible with datadog?

pkg/config/config.go
marceloboeira: sure, I forgot to regenerate the gopkg file, will do!

marceloboeira:
@zhouzhuojie it is, but it doesn't have all the features, e.g.: tags

pkg/config/middleware.go
marceloboeira: @zhouzhuojie I've been experimenting, the Datadog statsd supports tags, but this one I've used doesn't.

I might switch to the Datadog one, considering its backwards compatible with any other statsd whilst the normal statsd ones have missing features comparing to the Datadog one.

marceloboeira: in regards to the routes itself, the biggest issue if we include on the metric name is that it gets tricky to remove the "ids", which then becomes a bit useless.

you don't want to track metrics for resources, yet, routes:

`/api/foo/bar/1`  has to become `/api/foo/bar/` to be useful.

marceloboeira:
And another thing is to add such metrics to individual parts of the code, such as, calling database/kafka calls, so we can track down any possible latency or spikes. But I would probably cover that on another PR.

marceloboeira:
@zhouzhuojie updated.

- Using Datadog client (supports tags and it is backwards compatible with other statsd)
- Use tags for http status, url and method (allows better filtering later on, e.g.: duration of gets to /foo/bar)

marceloboeira:
I was considering moving the middlewares to sub-packages on their own files, since the file is getting a bit big and there is a lot of logic that would be better handled on individual files. But, I probably will do that on another PR.

Gopkg.lock
```diff
@@ -1,6 +1,12 @@
 # This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
 
 
+[[projects]]
```
zhouzhuojie: can you check in Gopkg.toml as well? this lock file should be derived from `Gopkg.toml`.

zhouzhuojie:
Thanks @marceloboeira! Left one comment, and I think it's good to go!

marceloboeira:
@zhouzhuojie yeap , no problem.

marceloboeira:
@zhouzhuojie done. 

I don't know why the `dep ensure` didn't work the first time.

btw, I started using the statsd metrics on a couple of pods running in production yesterday, pretty neat:

screen shot 2018-04-12 at 10 22 22

.gitattributes
```diff
@@ -0,0 +1,5 @@
+# All files use LF as eol
+* text=auto eol=lf
+
+# Remove vendor folder from the diff
+vendor/**/* -diff
```
zhouzhuojie: Good to learn about it!

PR #116, opened 2018-04-23
Fix test warning

zhouzhuojie:
- Fix test warning from swagger validation
- Use `go test` for test caching natively


docs/api_docs/bundle.yaml
```diff
@@ -764,7 +764,6 @@ definitions:
         type: integer
         format: int64
         minimum: 1
-        readOnly: true
```
zhouzhuojie: here `id` is required, setting it as `readOnly` will result in a warning in validation

Issue #115, opened 2018-04-11
Respect log level


marceloboeira:
I want to avoid overflowing my logs with:

```
{"FlagEvalResult":{"evalContext":{"entityID":"a8f155220fad675898b52160340a85e2","entityType":"user","flagID":1},"evalDebugLog":{"segmentDebugLogs":[]},"flagID":1,"flagSnapshotID":36,"segmentID":1,"timestamp":"2018-04-11T10:21:17Z","variantAttachment":{},"variantID":2,"variantKey":"heart"}}
{"FlagEvalResult":{"evalContext":{"entityID":"ea132b7d85d5fc333efda94088923f44","entityType":"user","flagID":1},"evalDebugLog":{"segmentDebugLogs":[]},"flagID":1,"flagSnapshotID":36,"segmentID":1,"timestamp":"2018-04-11T10:21:18Z","variantAttachment":{},"variantID":2,"variantKey":"heart"}}
{"FlagEvalResult":{"evalContext":{"entityID":"02f026d0f04d03818f5d41f3b73f062b","entityType":"user","flagID":1},"evalDebugLog":{"segmentDebugLogs":[]},"flagID":1,"flagSnapshotID":36,"segmentID":1,"timestamp":"2018-04-11T10:21:19Z","variantAttachment":{},"variantID":2,"variantKey":"heart"}}
{"FlagEvalResult":{"evalContext":{"entityID":"fa8f1f5ccbafcae6fe40f4d752019a1c","entityType":"user","flagID":1},"evalDebugLog":{"segmentDebugLogs":[]},"flagID":1,"flagSnapshotID":36,"segmentID":1,"timestamp":"2018-04-11T10:21:21Z","variantAttachment":{},"variantID":2,"variantKey":"heart"}}
{"FlagEvalResult":{"evalContext":{"entityID":"50f18fc583ed933f924256d28912cf26","entityType":"user","flagID":1},"evalDebugLog":{"segmentDebugLogs":[]},"flagID":1,"flagSnapshotID":36,"segmentID":1,"timestamp":"2018-04-11T10:21:22Z","variantAttachment":{},"variantID":1,"variantKey":"star"}}
[...]

PR #117, opened 2018-04-23
Use env to config eval logging

zhouzhuojie:
Use an env variable `FLAGR_EVAL_LOGGING_ENABLED` to control if we should log the eval result to the console or not.


zhouzhuojie:
fix #115

PR #118, opened 2018-05-01
Use random entity id if it's nil

zhouzhuojie:
## Description
If `entity_id` is null or empty, flagr will randomly generate one.

## Motivation and Context
Reduce the number of unexpected 422 errors. And also it's convenient for the use case that entity_id is not readily available

## How Has This Been Tested?
Unit test and local integration tests.

## Types of changes

### DIFF
--- a/pkg/handler/eval.go
+++ b/pkg/handler/eval.go
@@ -3,6 +3,7 @@ package handler
 import (
 	"encoding/json"
 	"fmt"
+	"math/rand"
 	"time"
 
 	"github.com/checkr/flagr/pkg/config"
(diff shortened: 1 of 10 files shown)


pkg/handler/eval.go
```diff
@@ -106,6 +107,10 @@ var evalFlag = func(evalContext models.EvalContext) *models.EvalResult {
 		return BlankResult(f, evalContext, fmt.Sprintf("flagID %v has no segments", flagID))
 	}
 
+	if evalContext.EntityID == "" {
+		evalContext.EntityID = fmt.Sprintf("randomly_generated_%d", rand.Int31())
```
jasondoc3: 👍

PR #122, opened 2018-05-10
Add db connection debugging logging env

zhouzhuojie:
Since the database connection error may contain credentials, we are adding this env to control whether we should log that error or not.

### DIFF
--- a/pkg/repo/db.go
+++ b/pkg/repo/db.go
@@ -23,7 +23,11 @@ func GetDB() *gorm.DB {
 	singletonOnce.Do(func() {
 		db, err := gorm.Open(config.Config.DBDriver, config.Config.DBConnectionStr)
 		if err != nil {
-			logrus.WithField("err", err).Fatal("failed to connect to db")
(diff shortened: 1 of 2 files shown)


pkg/repo/db.go
```diff
@@ -22,7 +22,7 @@ var (
 func GetDB() *gorm.DB {
 	singletonOnce.Do(func() {
 		db, err := gorm.Open(config.Config.DBDriver, config.Config.DBConnectionStr)
-		if err != nil {
+		if err != nil && config.Config.DBConnectionDebug {
 			logrus.WithField("err", err).Fatal("failed to connect to db")
 		}
```
ziru: It seems to me that you want `config.Config.DBConnectionDebug` to gate the output of `err` field, instead of the whole log message.

Is the change below what you intended to make?
```
if err != nil {
  if config.Config.DBConnectionDebug {
    logrus.WithField("err", err).Fatal("failed to connect to db")
  } else {
    logrus.Fatal("failed to connect to db")
  }
}
```

Issue #100, opened 2018-01-29
Roadmap


kalebdf:
Hi @zhouzhuojie,
This is a fantastic project! Do you & the Checkr team have a roadmap or vision for this project?

In no particular order, here are some features I've used in similar projects:
- GET All Flags - Limit, Pagination, Filter (by active/inactive, by tag, by namespace, etc)
- Tagging of Flags (for filtering sake, different apps)
- Namespaces of Flags ([Planout](https://facebook.github.io/planout/docs/namespaces.html))
- Record forcing user into variation bucket (future calls requesting variation for user will return forced variation)

zhouzhuojie:
Thank you! These are awesome feature requests! We have plans on the data metrics side, and if it's necessary, I can work on the features like search flags, flag tags during our next hack days. 

Namespace (i.e. layers) is awesome! I've seen implementation before in other systems, and typically it's not needed all the time, I will probably defer that to future major releases.

Record forcing users into variation bucket or Sticky rollout looks good, but I don't have a clear spec in mind. Right now flagr uses EntityID to hash users into variation bucket. If you don't change the rollout percent or distribution, the same user should always be bucketed into the same variant.

PRs are always welcome! 🎉

kalebdf:
@zhouzhuojie Stellar! We currently are sending all of our `Experiment Viewed` metrics into Segment which flows into our data warehouse / data lake. 🏢  📊 

I looked into `go-swagger` briefly to try to add filtering, limiting, & pagination, but didn't find it. I'm sure they have it though.

The use case for recording user (ID: 123) forced into variation bucket (ID: XYZ) and then having that stick is for the fallback case. Upon service timeout, we can default the feature experience for the user to variation (ID: XYZ). It's not high up on my list at all. I only noticed it on SeatGeek's [Sixpack](http://sixpack.seatgeek.com).

If we build upon **Flagr**, we'll definitely submit a PR. Thanks! 👍

nitin-vats:
I want to know how you decide which variant is the winner in A/B testing? Is there anything else which I need to integrate too?

PR #126, opened 2018-06-06
add support for querying flags by description, enabled and with limit

amalfra:
## Description

This PR adds support for querying flags by description and enabled status; also adds support for limiting number of results returned using limit query parameter

## Motivation and Context

https://github.com/checkr/flagr/issues/100 - the flags endpoint currently does not support filtering and limiting results. this seems to be useful feature for clients consuming the API.

## Types of changes

## Checklist:


pkg/handler/crud.go
```diff
@@ -68,6 +70,34 @@ var (
 func (c *crud) FindFlags(params flag.FindFlagsParams) middleware.Responder {
 	fs := []entity.Flag{}
 	q := entity.NewFlagQuerySet(getDB())
+	queryParams := map[string]string{
```
zhouzhuojie: I think it's better to put the query params in swagger query params.

https://github.com/checkr/flagr/blob/master/swagger/flags.yaml#L1

For example, https://swagger.io/docs/specification/describing-parameters/#query-parameters

After the change of the swagger definitions, run `make gen` to get the latest swagger model. We can get type safety and API doc for free in this case.

amalfra:
@zhouzhuojie I have updated the pr to use swagger.yml for query param validation

kalebdf:
Awesome work @amalfra 👍

PR #127, opened 2018-06-11
UI and API changes to support named flags (#1)

kenli-rubrik:
* Update model, get and create API to support unique flag names
* Support flag names in UI
* Support querying by name for flags
* Add test and update demo DB

(Merging from scaledata/flagr which was already reviewed)


kenli-rubrik:
Oops, didn't mean to submit this to checkr/flagr master.

Issue #133, opened 2018-06-23
getting error when running make build


amalfra:
After getting latest code(doing dep ensure, make deps etc) and running make gen, make build throws following error:
```
# github.com/checkr/flagr/swagger_gen/restapi
swagger_gen/restapi/server.go:174:18: cannot use domainSocket (type *graceful.Server) as type *http.Server in argument to configureServer
swagger_gen/restapi/server.go:206:18: cannot use httpServer (type *graceful.Server) as type *http.Server in argument to configureServer
swagger_gen/restapi/server.go:292:18: cannot use httpsServer (type *graceful.Server) as type *http.Server in argument to configureServer
make: *** [build] Error 2
```

## Expected Behavior

Generated binary sucessfully

## Current Behavior

Errors, can't compile

## Steps to Reproduce (for bugs)

1. pull latest code
2. do dep ensure
3. do make deps
4. do make gen
5. do make build

## Context

I am trying to add new query parameter for rest api but error prevents me from successfully compiling

## Your Environment

* Version used (`flagr version`):
* Server type and version:
* Operating System and version (`uname -a`):
* Link to your project:

zhouzhuojie:
Related to https://github.com/go-swagger/go-swagger/pull/1574

amalfra:
@zhouzhuojie thanks for replaying. I tried go get, make all, make deps etc. before doing make build. Still I get the error. I think my dependencies are messed up. Let me try if I can clean all vendor packages and see if it works

zhouzhuojie:
Landed two prs, let me know if it works

amalfra:
but now I get this package error when doing ```make all```. do I need to run any other commands before that?
```
Check deps
Checking project is in GOPATH
Installing retool
retool: installing github.com/golang/dep/cmd/dep
retool: installing github.com/alecthomas/gometalinter
retool: installing github.com/go-swagger/go-swagger/cmd/swagger
retool: installing github.com/jirfag/go-queryset/cmd/goqueryset
retool: installing github.com/dave/courtney
retool: installing github.com/codeskyblue/fswatch
Installing:
  deadcode
  dupl
  errcheck
  gas
  gochecknoglobals
  gochecknoinits
  goconst
  gocyclo
  goimports
  golint
  gosimple
  gotype
  gotypex
  ineffassign
  interfacer
  lll
  maligned
  megacheck
  misspell
  nakedret
  safesql
  staticcheck
  structcheck
  unconvert
  unparam
  unused
[...]

PR #136, opened 2018-06-26
Fix sed so it works for both mac and gnu sed

zhouzhuojie:
`sed -i ''` won't work for gnu sed, this is a fix that works for both.

https://stackoverflow.com/a/22084103
related to #133


zhouzhuojie:
Tested on a Dockerfile-CI container for the linux sed

PR #140, opened 2018-07-06
Add health handler

zhouzhuojie:
Tested locally for the health endpoint

```
curl localhost:18000/api/v1/health 
```


zhouzhuojie:
Fixes #102

marceloboeira:
👍

mukesh-p:
I was setting this health-check for a pod in GKE as a readiness probe. By default when I hit the server with curl from localhost on the path given it returns with "Not authorized" response. When I passed authorization headers along with curl then it worked. The problem is these authorization headers are not ideal for GKE HC probes and Ingress is always marked as unhealthy. Do we have some URL paths that gives back a 200 OK response without any need to authorize? I believe for HC probes it should be as simple as possible to get know about the server health.

zhouzhuojie:
@mukesh-p did you turn on the `FLAGR_BASIC_AUTH_ENABLED`? If so, you may also want to whitelist the `/api/v1/health` endpoint. 

You can set 
```
# Default
FLAGR_BASIC_AUTH_WHITELIST_PATHS=/api/v1/flags,/api/v1/evaluation

# You can add /api/v1/health or change it
FLAGR_BASIC_AUTH_WHITELIST_PATHS=/api/v1/flags,/api/v1/evaluation,/api/v1/health
``` 

I'm also thinking of adding `/api/v1/health` to the default.

zhouzhuojie:
Related https://github.com/checkr/flagr/pull/407

PR #144, opened 2018-07-10
Add test coverage for middleware

zhouzhuojie:


pkg/config/middleware_test.go
```diff
@@ -28,3 +34,95 @@ func TestSetupGlobalMiddleware(t *testing.T) {
...
+
+		res := httptest.NewRecorder()
+		res.Body = new(bytes.Buffer)
+		req, _ := http.NewRequest("GET", "http://localhost:18000/api/v1/flags", nil)
+		req.AddCookie(&http.Cookie{
+			Name:  "access_token",
+			Value: "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJmbGFncl91c2VyIjoiMTIzNDU2Nzg5MCJ9.CLXgNEtwPCqCOtUU-KmqDyO8S2wC_G6PZ0tml8DCuNw", // {"flagr_user": "1234567890"}
+		})
+		hh.ServeHTTP(res, req)
+		assert.NotEqual(t, 307, res.Code)
```
rstratton: Nit, might be better to `assert.Equal` the actual value we get, so that tests document behavior more directly.

pkg/config/middleware_test.go
```diff
@@ -28,3 +34,95 @@ func TestSetupGlobalMiddleware(t *testing.T) {
...
+	t.Run("it will setup statsd if statsd is enabled", func(t *testing.T) {
+		Config.StatsdEnabled = true
+		hh = SetupGlobalMiddleware(h)
+
+		res := httptest.NewRecorder()
+		res.Body = new(bytes.Buffer)
+		req, _ := http.NewRequest("GET", "http://localhost:18000/api/v1/flags", nil)
+
+		incrCalled := false
+		monkey.PatchInstanceMethod(
```
rstratton: 🐵

zhouzhuojie: Yup, I was mindblown by this testing package `monkey`, it's so powerful that I don't even need to mock the functions. It replaces a function at runtime.

https://bou.ke/blog/monkey-patching-in-go/

rstratton: That's crazy, so cool!

PR #145, opened 2018-07-10
Add coverage for kafka data recorder

zhouzhuojie:


pkg/handler/data_recorder_kafka.go
```diff
@@ -143,7 +143,7 @@ func (r *kafkaEvalResult) Length() int {
 
 // Key generates the partition key
 func (r *kafkaEvalResult) Key() string {
-	if r.EvalResult == nil {
+	if r.EvalResult == nil || r.EvalContext == nil {
```
rstratton: What does this change do?

zhouzhuojie: an edge case found via the unit test. The line below `r.EvalContext.EntityID` requires that r.EvalContext is not nil

PR #146, opened 2018-07-20
Add JWT auth via headers and RS256 signing option

vayan:
## Description

It's more common to use JWT tokens via headers. (e.g: `Authorization Bearer token11`)

Also added the RS256 signing method which is used a lot too.

By default everything works like before 👍 

## How Has This Been Tested?

Manual testing + lot of new UT 

## Types of changes

## Checklist:

### DIFF
--- a/pkg/config/env.go
+++ b/pkg/config/env.go
@@ -90,4 +90,5 @@ var Config = struct {
 	JWTAuthSecret             string   `env:"FLAGR_JWT_AUTH_SECRET" envDefault:""`
 	JWTAuthNoTokenRedirectURL string   `env:"FLAGR_JWT_AUTH_NO_TOKEN_REDIRECT_URL" envDefault:""`
 	JWTAuthUserProperty       string   `env:"FLAGR_JWT_AUTH_USER_PROPERTY" envDefault:"flagr_user"`
+	JWTAuthSigningMethod      string   `env:"FLAGR_JWT_AUTH_SIGNING_METHOD" envDefault:"HS256"`
 }{}
--- a/pkg/config/env.go
+++ b/pkg/config/env.go
@@ -79,10 +79,24 @@ var Config = struct {
 	RecorderKafkaEncrypted      bool          `env:"FLAGR_RECORDER_KAFKA_ENCRYPTED" envDefault:"false"`
 	RecorderKafkaEncryptionKey  string        `env:"FLAGR_RECORDER_KAFKA_ENCRYPTION_KEY" envDefault:""`
 
-	// JWTAuthEnabled enables the JWT Auth
-	// The pattern of using JWT auth token is that it redirects to the URL to set cross subdomain cookie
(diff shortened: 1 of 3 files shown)


vayan:
Linter is different on the CI? 

`gometalinter --config=.gometalinter.json ./pkg/...` returns no error locally

pkg/config/env.go
```diff
@@ -90,4 +90,5 @@ var Config = struct {
 	JWTAuthSecret             string   `env:"FLAGR_JWT_AUTH_SECRET" envDefault:""`
 	JWTAuthNoTokenRedirectURL string   `env:"FLAGR_JWT_AUTH_NO_TOKEN_REDIRECT_URL" envDefault:""`
 	JWTAuthUserProperty       string   `env:"FLAGR_JWT_AUTH_USER_PROPERTY" envDefault:"flagr_user"`
+	JWTAuthSigningMethod      string   `env:"FLAGR_JWT_AUTH_SIGNING_METHOD" envDefault:"HS256"`
```
zhouzhuojie: Let's add a comment for possible values of signing methods.

The whole file will be served as the doc at https://checkr.github.io/flagr/#/flagr_env

zhouzhuojie: Also, can you help change the comments above of the pattern of how to use JWT auth?

pkg/config/middleware_test.go
```diff
@@ -100,6 +117,120 @@ func TestAuthMiddleware(t *testing.T) {
...
+		req, _ := http.NewRequest("GET", "http://localhost:18000/api/v1/flags", nil)
+		req.AddCookie(&http.Cookie{
+			Name:  "access_token",
+			Value: validHS256JWTTokenWithSecret,
+		})
+		hh.ServeHTTP(res, req)
+		assert.Equal(t, http.StatusOK, res.Code)
+	})
+
+	t.Run("it will pass if jwt enabled with correct HS256 token cookie and signing method is wrong", func(t *testing.T) {
```
zhouzhuojie: nit: change to "it will pass with a correct HS256 token cookie when signing method is wrong and it defaults to empty string secret"

zhouzhuojie:
Thanks for the PR! Welcome to Flagr! Want to collect some feedback on how are you going to use Flagr, for example, are you building your own UI or simply run it through some generated clients?

Changing to RS256 and sending through headers also requires some changes to the Flagr UI, but I guess you can put it in another pr.

pkg/config/env.go
vayan: Sure! I missed this comment 👍

vayan:
> Changing to RS256 and sending through headers also requires some changes to the Flagr UI, but I guess you can put it in another pr.

What changes need to be done? 

>  how are you going to use Flagr

I'm going to use Flagr using https://github.com/checkr/jsflagr on our backend and manual changes via the UI.

Issue #150, opened 2018-08-01
Kinesis Data Recorder Adapter


marceloboeira:
At my company we use mostly Kinesis for streaming, so I would like to start working on a kinesis recorder; @zhouzhuojie wdyt?

zhouzhuojie:
That's awesome, I've already reserved the switch case for more data recorders.

https://github.com/checkr/flagr/blob/master/pkg/handler/data_recorder.go#L24

marceloboeira:
Closing since #151 got merged!

PR #151, opened 2018-08-04
Add Kinesis support

marceloboeira:
From: https://github.com/checkr/flagr/issues/150

## Description

Add Kinesis support for the data recorder.
  

## Motivation and Context

 That makes it more flexible since Kafka is not easy to self-host and Kinesis works well-enough for small throughput rates.

## How Has This Been Tested?

Still figuring how we should test this since Kinesis is not open source, mocking? or just best effort (unit + offline integration?)
 

## Types of changes

## Checklist:

- [x] Delivery Test

##  Delivery Test

For the delivery test I've created a small script to trigger 10k requests, to the evaluation endpoint, in parallel (batches of 8/10). That should `put` at least 10k records into the stream.

e.g.: 
```script
seq 1 10000 | parallel -I {} http --ignore-stdin post 'http://127.0.0.1:18000/api/v1/evaluation'  entityID=neu-{} entityType=report flagID:=3 > neu.txt
``` 

In the meantime I was reading from the stream and storing the content into a file. Later on I've grep the content of the file, looking for the IDs of the entities sent, filtering that out, sorting, checking if its uniq and counting the uniq IDs.

`grep "neu-[0-9]*" -ao kinesis.log | sort -V | uniq -c | sort`

That returned 10k.


pkg/handler/data_recorder_test.go
```diff
@@ -1,15 +1,39 @@
...
+
 	assert.NotPanics(t, func() {
 		GetDataRecorder()
 	})
 }
+
+func TestGetDataRecorderWhenKinesis(t *testing.T) {
+	singletonDataRecorderOnce = sync.Once{}
+	defer gostub.StubFunc(&NewKinesisRecorder, nil).Reset()
+	config.Config.RecorderType = "kinesis"
```
zhouzhuojie: Let's also set the RecorderType back to default after this test

pkg/handler/data_recorder_kinesis.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		logrus.WithField("kinesis_error", err).Error("error pushing")
+	}
+}
+
+type kinesisEvalResult struct {
+	*models.EvalResult
+}
+
+// Unmalshalls the context
```
zhouzhuojie: nit, comment looks wrong. Do you mean `Payload marshals the EvalResult`?

pkg/handler/data_recorder_kinesis.go
```diff
@@ -0,0 +1,87 @@
...
+	kr := &kinesisEvalResult{
+		EvalResult: r,
+	}
+
+	payload, err := kr.Payload()
+	if err != nil {
+		logrus.WithField("kinesis_error", err).Error("error marshaling")
+	}
+
+	err = k.producer.Put(payload, kr.Key())
```
zhouzhuojie: Currently the kafka record schema looks like

```
{
  "payload": string(EvalResult.MarshalBinary()) OR encrypted string,
  "encrypted": false   <---- optional, you can ignore this and implement encryption later if needed
}
```

Can we make the same abstraction/schema in kinesis? So that it's seamless to switch data recorders.

Another idea is to move https://github.com/checkr/flagr/blob/master/pkg/handler/data_recorder_kafka.go#L152 as the shared "message frame" for both Kafka and Kinesis

pkg/handler/data_recorder_kinesis.go
```diff
@@ -0,0 +1,87 @@
...
+		AggregateBatchSize:  config.Config.RecorderKinesisAggregateBatchSize,
+		Verbose:             config.Config.RecorderKinesisVerbose,
+		Logger:              logrus.WithField("producer", "kinesis"),
+	})
+
+	p.Start()
+
+	go func() {
+		for err := range p.NotifyFailures() {
+			logrus.WithField("kinesis_error", err).Error("error pushing")
```
zhouzhuojie: nit, the error message can be more specific, like `logrus.WithField("kinesis_error", err).Error("error pushing to kinesis")`, same for the error messages below

pkg/handler/data_recorder_kinesis_test.go
```diff
@@ -0,0 +1,39 @@
+package handler
+
+import (
+	"testing"
+
+	"github.com/checkr/flagr/pkg/util"
+	"github.com/checkr/flagr/swagger_gen/models"
+	//"github.com/prashantv/gostub"
```
zhouzhuojie: remove comments

pkg/config/env.go
```diff
@@ -79,6 +79,17 @@ var Config = struct {
 	RecorderKafkaEncrypted      bool          `env:"FLAGR_RECORDER_KAFKA_ENCRYPTED" envDefault:"false"`
 	RecorderKafkaEncryptionKey  string        `env:"FLAGR_RECORDER_KAFKA_ENCRYPTION_KEY" envDefault:""`
 
+	// Kinesis related configurations for data records logging (Flagr Metrics)
+	RecorderKinesisStreamName          string        `env:"FLAGR_RECORDER_KINESIS_STREAM_NAME" envDefault:"flagr-stream"`
```
zhouzhuojie: can be the same default as kafka, i.e. `flagr-records`

pkg/config/env.go
```diff
@@ -79,6 +79,17 @@ var Config = struct {
...
+	RecorderKinesisBacklogCount        int           `env:"FLAGR_RECORDER_KINESIS_BACKLOG_COUNT" envDefault:"500"`
+	RecorderKinesisMaxConnections      int           `env:"FLAGR_RECORDER_KINESIS_MAX_CONNECTIONS" envDefault:"24"`
+	RecorderKinesisFlushInterval       time.Duration `env:"FLAGR_RECORDER_KINESIS_FLUSH_INTERVAL" envDefault:"5s"`
+	RecorderKinesisBatchCount          int           `env:"FLAGR_RECORDER_KINESIS_BATCH_COUNT" envDefault:"500"`
+	RecorderKinesisBatchSize           int           `env:"FLAGR_RECORDER_KINESIS_BATCH_SIZE" envDefault:"0"`
+	RecorderKinesisAggregateBatchCount int           `env:"FLAGR_RECORDER_KINESIS_AGGREGATE_BATCH_COUNT" envDefault:"4294967295"`
+	RecorderKinesisAggregateBatchSize  int           `env:"FLAGR_RECORDER_KINESIS_AGGREGATE_BATCH_SIZE" envDefault:"51200"`
+	RecorderKinesisVerbose             bool          `env:"FLAGR_RECORDER_KINESIS_VERBOSE" envDefault:"false"`
+
```
zhouzhuojie: Also, what about kinesis connection credentials, not very familiar with kinesis, I would image you may need AWS keys.

pkg/handler/data_recorder_kinesis.go
marceloboeira: 👍

pkg/handler/data_recorder_kinesis_test.go
marceloboeira: 👍

pkg/handler/data_recorder_kinesis.go
marceloboeira: 👍

pkg/handler/data_recorder_test.go
marceloboeira: 👍

pkg/config/env.go
marceloboeira: Yes, you need, but they are loaded implicitly by default from AWS packages, we can document but I wouldn't change the behavior of that.

pkg/config/env.go
marceloboeira: 👍

pkg/handler/data_recorder_kinesis.go
marceloboeira: Let's start with the duplication, and then I can open a second PR with the abstractions, so first I can implement and figure what will be duplicated.

pkg/config/env.go
zhouzhuojie: 👍 it would be great if we can have some docs that point to the AWS behavior.

pkg/handler/data_recorder_kinesis.go
zhouzhuojie: Sounds good for deferring the abstractions later. Let's at least make sure the message schema is the same, otherwise, there will be breaking changes for Kinesis in the future.

marceloboeira: okay 👍

pkg/config/env.go
marceloboeira: okay 👍

pkg/handler/data_recorder_kinesis.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+type kinesisRecorder struct {
+	enabled  bool
+	producer *producer.Producer
+}
+
+// NewKinesisRecorder creates a new Kinesis recorder
+var NewKinesisRecorder = func() DataRecorder {
```
zhouzhuojie: it would be great if we can cover more lines

https://codecov.io/gh/checkr/flagr/pull/151/diff?src=pr&el=tree#diff-cGtnL2hhbmRsZXIvZGF0YV9yZWNvcmRlcl9raW5lc2lzLmdv

marceloboeira: Yes, it's still WIP, I have to test more things. I just didn't have time yet.

marceloboeira:
@zhouzhuojie I think the code its pretty much done, please take a look.  I'll be conducting some performance/delivery tests now to make sure like if I call 1M times, I have 1M records on the stream...

Once I have that ready I'll mention here so we move forward. In the meantime, please feel free to look at the code and I can work on your comments right away.

marceloboeira:
## Performance / Delivery Test

For the performance/delivery test I've created a small script to trigger 10k requests, to the evaluation endpoint, in parallel (batches of 8/10). That should `put` at least 10k records into the stream.

e.g.: 
```script
seq 1 10000 | parallel -I {} http --ignore-stdin post 'http://127.0.0.1:18000/api/v1/evaluation'  entityID=neu-{} entityType=report flagID:=3 > neu.txt
``` 

In the meantime I was reading from the stream and storing the content into a file. Later on I've grep the content of the file, looking for the IDs of the entities sent, filtering that out, sorting, checking if its uniq and counting the uniq IDs.

`grep "neu-[0-9]*" -ao kinesis.log | sort -V | uniq -c | sort`

That returned 10k.

PR #170, opened 2018-09-25
Added config for base path for web UI

SebastianOsuna:
## Description
Added `FLAGR_WEB_PREFIX` env so it is possible to serve the static web UI from another path different to `/`

## Motivation and Context

For example, if deploying flagger in ECS behind and ELB it wouldn't be possible to deploy Flagr anywhere but in the `/*` listener since the Web UI is served in `/` and the API in `/api/v1`, this would allow to server the UI from `/web/flagr/*` for example.

## How Has This Been Tested?
Built the code in a docker image, set the environment variable to `/web/flagr` and the UI was loading from that prefix

## Types of changes

## Checklist:
- [?] My code follows the code style of this project.
- [X] I have updated the documentation accordingly.
- [?] All new and existing tests passed.

### DIFF
--- a/pkg/config/env.go
+++ b/pkg/config/env.go
@@ -120,4 +120,7 @@ var Config = struct {
 
 	// "HS256" and "RS256" supported
 	JWTAuthSigningMethod string `env:"FLAGR_JWT_AUTH_SIGNING_METHOD" envDefault:"HS256"`
+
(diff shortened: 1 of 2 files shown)


SebastianOsuna:
I'm unsure if I am following the code style or if this change needs a test, please let me know, or any other comments. Thanks

PR #171, opened 2018-10-03
Add entity_type override

zhouzhuojie:
Add entity_type to flag settings, so that it's convenient for users to override the entity_type in evaluation logs. 

## Motivation and Context
Sometimes entity_types are not normalized, and it makes hard for data pipeline to handle all different kinds of entity_types. This pr provides a backward compatible solution that people can override the entity_type when it's necessary. 

## How Has This Been Tested?
- tested locally
- tested with UI
- tested with evaluation logs
- tested with evaluation debugging

## Types of changes

## Checklist:


zhouzhuojie:
UI Testing

image


image


image


image

pkg/entity/db.go
```diff
@@ -18,6 +18,18 @@ var (
 	singletonOnce sync.Once
 )
 
+// AutoMigrateTables stores the entity tables that we can auto migrate in gorm
```
zhouzhuojie: Moved from pkg/entity/entity.go

pkg/entity/flag.go
```diff
@@ -95,3 +86,29 @@ func (qs FlagQuerySet) DescriptionLike(description string) FlagQuerySet {
...
+			return "", fmt.Errorf("cannot create flag due to invalid key. reason: %s", reason)
+		}
+	}
+	return key, nil
+}
+
+// CreateFlagEntityType creates the FlagEntityType if not exists
+func CreateFlagEntityType(db *gorm.DB, key string) error {
+	ok, reason := util.IsSafeKey(key)
+	if !ok && key != "" {
```
zhouzhuojie: Instead of just checking if the key is valid, I wonder if we can add an ENV variable to limit the possible values of entityTypes here.

pkg/handler/data_recorder_kinesis_test.go
```diff
@@ -63,8 +63,8 @@ func TestKinesisEvalResult(t *testing.T) {
 		assert.Equal(t, k, "d08042018")
 
 		p, err := r.Payload()
-		assert.Equal(t, err, nil)
-		assert.Equal(t, "{\"evalContext\":{\"entityID\":\"d08042018\",\"entityType\":null},\"flagID\":1,\"flagKey\":\"random_flag_key\",\"flagSnapshotID\":1,\"segmentID\":1,\"timestamp\":null,\"variantAttachment\":null,\"variantID\":1,\"variantKey\":\"control\"}", string(p))
+		assert.NoError(t, err)
+		assert.NotEmpty(t, p)
```
zhouzhuojie: I've made the tests not that brittle due to new fields in the evaluation results.

pkg/handler/eval_cache.go
```diff
@@ -66,13 +66,16 @@ var fetchAllFlags = func() ([]entity.Flag, error) {
...
 	// doc: http://jinzhu.me/gorm/crud.html#preloading-eager-loading
 	fs := []entity.Flag{}
-	err := getDB().Preload("Segments", func(db *gorm.DB) *gorm.DB {
-		return db.Preload("Distributions", func(db *gorm.DB) *gorm.DB {
-			return db.Order("variant_id ASC")
-		}).Preload("Constraints", func(db *gorm.DB) *gorm.DB {
-			return db.Order("created_at ASC")
-		}).Order("rank ASC").Order("id ASC")
-	}).Preload("Variants").Find(&fs).Error
+	err := getDB().
```
zhouzhuojie: Here's just some formatting

michberr:
nit: entity type is a little squished, would be nice if it lined up with the flag description cell
screen shot 2018-10-05 at 10 13 01 am

pkg/entity/db.go
michberr: 👍

pkg/entity/flag.go
```diff
@@ -15,15 +15,17 @@ import (
...
+	Description string `sql:"type:text"`
+	CreatedBy   string
+	UpdatedBy   string
+	Enabled     bool
+	Segments    []Segment
+	Variants    []Variant
+	SnapshotID  uint `json:"-"`
+
 	DataRecordsEnabled bool
-	SnapshotID         uint `json:"-"`
```
michberr: What was this field used for?

pkg/entity/flag.go
michberr: I'm not sure that an ENV variable will add more standardization/is less prone to error than adding in the flagr UI - seems like just one more annoying step. However, this one field does have the potential to affect analytics in real-time and could make things very wonky

pkg/handler/data_recorder_kinesis_test.go
michberr: how about a test with regex to match some important pieces of the payload like entityID, flagKey etc

pkg/entity/flag.go
zhouzhuojie: Oh, I didn't remove it, just moved it a little bit above. It was used to store the FlagSnapshotID

docs/api_docs/bundle.yaml
```diff
@@ -132,48 +132,49 @@ paths:
...
-          schema:
-            $ref: '#/definitions/flag'
+          description: OK deleted
         default:
           description: generic error response
           schema:
             $ref: '#/definitions/error'
-    delete:
+    put:
+      deprecated: true
```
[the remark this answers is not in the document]
zhouzhuojie: Good catch, originally I was planning to deprecate put and use patch, but it's confusing and I may just have to stick to put.

pkg/entity/flag.go
zhouzhuojie: Maybe we can fork the open source UI and create a private one just for Checkr.

zhouzhuojie:
> nit: entity type is a little squished, would be nice if it lined up with the flag description cell
> screen shot 2018-10-05 at 10 13 01 am



Good catch, I've changed to use two rows.

image

pkg/entity/flag.go
zhouzhuojie: As discussed, we will make it a UI env in the next pr

PR #172, opened 2018-10-06
Allow uppercase letters in flag keys

kalebdf:
Flag keys may now allow uppercase letters.

## Description
This PR allows Flag keys to be alphanumeric & case insensitive as well as include underscores. It does remove the constraint of having at least one lowercase letter as the first character.

## Motivation and Context

The first FLAG in the [Flagr demo](https://try-flagr.herokuapp.com) had a flag key (e.g. `kppaP9af9x10`) which had an uppercase letter which threw an error when I tried to change the description because of the capital letters in the flag key. 

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -12,9 +12,8 @@ import (
 
 var (
 	keyLengthLimit = 63
-	keyRegex       = regexp.MustCompile("^[a-z]+[a-z0-9_]*$")
+	keyRegex       = regexp.MustCompile("(?i)^[a-z0-9_]+$")
 
-	randomKeyCharset = []byte("123456789abcdefghijkmnopqrstuvwxyz")
 	randomKeyPrefix  = "k"
 )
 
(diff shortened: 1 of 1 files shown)


zhouzhuojie:
Thanks for the PR, I think I will fix the demo sqlite db, not the valid key regex.

zhouzhuojie:
The demo db was fixed in https://github.com/checkr/flagr/pull/173. 
So https://try-flagr.herokuapp.com should be good to go now.

Yes, you can just run `/export/sqlite`, or start a new one with env `DBConnectionStr` points to it, it's basically a binary file.

> Also, is there a reason for only allowing lowercase characters + numbers + underscores in keys?

I don't have a strong reason, it was used to limit the charset of variant key at the beginning. Keys can be changed by users (e.g. Variant Key, Flag Key), and without uppercase it's just easier to search.

kalebdf:
Makes sense. Thanks @zhouzhoujie.

PR #175, opened 2018-10-08
Add UI env for entity types override

zhouzhuojie:
## Description
Use `FLAGR_UI_POSSIBLE_ENTITY_TYPES` as build arg for UI, by default it's `null`, which means it doesn't imply any restrictions.

## Motivation and Context
To limit the choices and normalize entity types for better data analytics

## How Has This Been Tested?
Tested locally and with docker build-arg

## Types of changes

### DIFF
--- a/browser/flagr-ui/src/components/Flag.vue
+++ b/browser/flagr-ui/src/components/Flag.vue
@@ -176,7 +176,7 @@
                           v-model="flag.entityType"
                           size="mini"
                           filterable
-                          allow-create
+                          :allow-create="allowCreateEntityType"
                           default-first-option
                           placeholder="<null>">
                           <el-option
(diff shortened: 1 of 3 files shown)


browser/flagr-ui/src/components/Flag.vue
```diff
@@ -774,16 +775,29 @@ export default {
...
         })
-        if (response.data.indexOf('') === -1) {
+        if (entityTypes.indexOf('') === -1) {
           arr.unshift({label: '<null>', value: ''})
         }
-        this.entityTypes = arr
+        return arr
+      }
+
+      if (process.env.FLAGR_UI_POSSIBLE_ENTITY_TYPES) {
```
michberr: FLAGR_UI_POSSIBLE_ENTITY_TYPES

zhouzhuojie: It's a bit different, I have to use `process.env.FLAGR_UI_POSSIBLE_ENTITY_TYPES` here, and also it also needs to be registered in `browser/flagr-ui/config`. What the compiler does is to do a string interpolation by using the env variables.

PR #177, opened 2018-10-09
Add dd metrics

zhouzhuojie:
Add datadog metrics for

- Flag snapshot updated events
- Evaluation

### DIFF
--- a/pkg/entity/flag_snapshot.go
+++ b/pkg/entity/flag_snapshot.go
@@ -4,7 +4,10 @@ package entity
 
 import (
 	"encoding/json"
+	"fmt"
 
+	"github.com/checkr/flagr/pkg/config"
+	"github.com/checkr/flagr/pkg/util"
 	"github.com/jinzhu/gorm"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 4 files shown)


pkg/entity/flag_snapshot.go
```diff
@@ -66,4 +68,21 @@ func SaveFlagSnapshot(db *gorm.DB, flagID uint, updatedBy string) {
...
+	logFlagSnapshotUpdate(flagID, updatedBy)
+}
+
+var logFlagSnapshotUpdate = func(flagID uint, updatedBy string) {
+	if config.Global.StatsdClient == nil {
+		return
+	}
+
+	config.Global.StatsdClient.Incr(
+		"flag.snapshot.updated",
```
michberr: This will be namespaced by `flagr` right?

zhouzhuojie: yes

pkg/entity/flag_snapshot.go
```diff
@@ -66,4 +69,21 @@ func SaveFlagSnapshot(db *gorm.DB, flagID uint, updatedBy string) {
...
+		[]string{
+			fmt.Sprintf("FlagID:%d", flagID),
+			fmt.Sprintf("UpdatedBy:%s", util.SafeStringWithDefault(updatedBy, "null")),
```
ziru: what are the values range for updatedBy tag?
we should use relative bucket value (such as 1d-ago, 2d-ago), rather than absolute value?

pkg/handler/eval.go
```diff
@@ -169,13 +169,34 @@ var logEvalResult = func(r *models.EvalResult, dataRecordsEnabled bool) {
...
+	if config.Global.StatsdClient == nil {
+		return
+	}
+
+	config.Global.StatsdClient.Incr(
+		"evaluation",
+		[]string{
+			fmt.Sprintf("EntityType:%s", util.SafeStringWithDefault(r.EvalContext.EntityType, "null")),
+			fmt.Sprintf("FlagID:%d", util.SafeUint(r.FlagID)),
+			fmt.Sprintf("FlagSnapshotID:%d", util.SafeUint(r.FlagSnapshotID)),
```
ziru: is this necessary?
generally, we should use relative bucket value (such as 1d-ago, 2d-ago), rather than absolute value.

pkg/entity/flag_snapshot.go
zhouzhuojie: updatedBy is not a timestamp, it's the authenticated user

PR #178, opened 2018-10-10
Add kafka data recorder dd metrics

zhouzhuojie:


pkg/handler/data_recorder_kafka.go
```diff
@@ -106,6 +107,21 @@ func (k *kafkaRecorder) AsyncRecord(r *models.EvalResult) {
...
+}
+
+var logKafkaAsyncRecordToDatadog = func(r *models.EvalResult) {
+	if config.Global.StatsdClient == nil {
+		return
+	}
+	config.Global.StatsdClient.Incr(
+		"data_recorder.kafka",
+		[]string{
+			fmt.Sprintf("FlagID:%d", util.SafeUint(r.FlagID)),
```
ziru: nit: better use dash or underscores as delimiter.  since statsd tags will normalize to lowercase.

zhouzhuojie: yup, thanks for pointing it out. I just realized that in the dashboard, here just to keep it consistent because it was already used in other places.

PR #181, opened 2018-10-22
Fix segments evaluation

### LINKED ISSUE #180: Flagr segments evaluation should stop if it matches all the constraints in a segment
## Expected Behavior
If the entity context matches the constraints in a segment, it should stop evaluating the following segments.

Specifically, for 0% rollout segment, it shouldn't go to the next segments.

For example:

image

The first segment catches all the entity, and roll out to 0%, so every entity should have `null` variant returned.

## Current Behavior
It falls through to the next segment for non-rolled out users in the previous segments.

image

## Possible Solution
Stop evaluating the next segments if we found a match (i.e. satisfied all the constraints in a segment)

## Steps to Reproduce (for bugs)
https://try-flagr.herokuapp.com/#/flags/1

zhouzhuojie:
## Description
Fix #180 

This pr changed the behavior of segments fall-through logic. The intended behavior wad documented in https://checkr.github.io/flagr/#/flagr_use_cases

image

## How Has This Been Tested?
New multiple segments with 0% rollout flow.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/pkg/handler/eval.go
+++ b/pkg/handler/eval.go
@@ -140,13 +140,15 @@ var evalFlag = func(evalContext models.EvalContext) *models.EvalResult {
 	var sID *int64
 
 	for _, segment := range f.Segments {
-		variantID, log := evalSegment(f.ID, evalContext, segment)
+		sID = util.Int64Ptr(int64(segment.ID))
+		variantID, log, evalNextSegment := evalSegment(f.ID, evalContext, segment)
 		if evalContext.EnableDebug {
 			logs = append(logs, log)
 		}
 		if variantID != nil {
-			sID = util.Int64Ptr(int64(segment.ID))
 			vID = util.Int64Ptr(int64(*variantID))
+		}
+		if !evalNextSegment {
 			break
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/handler/eval_test.go
```diff
@@ -171,6 +176,26 @@ func TestEvalFlag(t *testing.T) {
 		assert.NotNil(t, result.VariantID)
 	})
 
+	t.Run("test multiple segments with the first segment 0% rollout", func(t *testing.T) {
+		f := entity.GenFixtureFlag()
+		f.Segments = append(f.Segments, entity.GenFixtureSegment())
+		f.Segments[0].Constraints = []entity.Constraint{}
+		f.Segments[0].RolloutPercent = uint(0)
```
zhouzhuojie: The first segment says `0%` rollout for all entity (no constraints), thus the evaluation result should have `nil` variantID.

pkg/handler/eval.go
```diff
@@ -220,14 +223,14 @@ var evalSegment = func(
 				Msg:       err.Error(),
 				SegmentID: int64(segment.ID),
 			}
-			return nil, log
+			return nil, log, true
```
rstratton: What kinds of errors do we expect from `conditions.Evaluate(expr, m)`?  Trying to understand why we should eval the next segment in this context

Issue #121, opened 2018-05-07
Question: Using the JWT Auth


mhaamann:
Do you have any recommendations on how to use JWT Auth with flagr. I was hoping that this could be handled with the Github OAuth. Is there any similar 3rd party vendors I could use?

PR #183, opened 2018-10-26
Make user claim configurable

zhouzhuojie:
Sometimes the user claim may not be `sub`, so here just make it easy to configure via ENV.


zhouzhuojie:
related to #121

PR #186, opened 2018-10-30
Add headers to CORS middleware

zhouzhuojie:


jaysonsantos:
Hey there, first thanks for the PR and the project
Second, it seems that this had no effect on CORS.
To reproduce it you can follow this:
```bash
make build run
# and then make an options call
curl -vXOPTIONS http://127.0.0.1:18000/api/v1/evaluation
```
and the output will be:

```
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 18000 (#0)
> OPTIONS /api/v1/evaluation HTTP/1.1
> Host: 127.0.0.1:18000
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 405 Method Not Allowed
< Allow: POST
< Content-Type: application/json
< Date: Mon, 20 Jan 2020 13:19:45 GMT
< Content-Length: 70
<
* Connection #0 to host 127.0.0.1 left intact
{"code":405,"message":"method OPTIONS is not allowed, but [POST] are"}* Closing connection 0
```
it seems that go-openapi's handler runs before negroni handlers and as on the spec there is no options method defined, it just cancel the request with 405.
What is the best approach to fix this? If you lead me I can try and send a PR for this.

PR #189, opened 2018-11-10
Add kafka version config

zhouzhuojie:
Add kafka version configuration


pkg/config/env.go
```diff
@@ -67,6 +67,7 @@ var Config = struct {
 	RecorderType string `env:"FLAGR_RECORDER_TYPE" envDefault:"kafka"`
 
 	// Kafka related configurations for data records logging (Flagr Metrics)
+	RecorderKafkaVersion        string        `env:"FLAGR_RECORDER_KAFKA_VERSION" envDefault:"0.8.2.0"`
```
zhouzhuojie: This is the default sarama setting

PR #190, opened 2018-11-12
Add integration tests

zhouzhuojie:
## Description
Add integration tests so that we can test flagr against multiple environments or multiple DBs.

## Motivation and Context
To cover DB discrepancies.

## How Has This Been Tested?
Will add more endpoints covered by the swagger API definition.

## Types of changes

## Checklist:

### DIFF
--- a/docs/api_docs/bundle.yaml
+++ b/docs/api_docs/bundle.yaml
@@ -4,7 +4,7 @@ info:
     Flagr is a feature flagging, A/B testing and dynamic configuration
     microservice. The base path for all the APIs is "/api/v1".
   title: Flagr
-  version: 1.0.12
+  version: 1.0.13
 tags:
   - name: flag
     description: Everything about the flag
@@ -986,7 +986,6 @@ definitions:
     type: object
     required:
       - key
-      - attachment
     properties:
       key:
         type: string
@@ -1060,9 +1059,6 @@ definitions:
         format: int64
         minimum: 0
         maximum: 100
-      bitmap:
-        type: string
-        minLength: 1
       variantKey:
         type: string
         minLength: 1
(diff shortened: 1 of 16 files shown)


docs/api_docs/bundle.yaml
```diff
@@ -986,7 +986,6 @@ definitions:
     type: object
     required:
       - key
-      - attachment
```
zhouzhuojie: no need to be required, tested with integration tests

docs/api_docs/bundle.yaml
```diff
@@ -1060,9 +1059,6 @@ definitions:
         format: int64
         minimum: 0
         maximum: 100
-      bitmap:
```
zhouzhuojie: not needed now, tested with integration tests

pkg/handler/crud.go
```diff
@@ -89,7 +89,7 @@ func (c *crud) FindFlags(params flag.FindFlagsParams) middleware.Responder {
 		q = q.Offset(int(*params.Offset))
 	}
 
-	err := q.All(&fs)
+	err := q.OrderAscByID().All(&fs)
```
zhouzhuojie: Need to explicitly sort by ID for postgres, caught by integration tests

pkg/handler/eval.go
```diff
@@ -213,7 +213,7 @@ var evalSegment = func(
 				Msg:       fmt.Sprintf("constraints are present in the segment_id %v, but got invalid entity_context: %s.", segment.ID, spew.Sdump(evalContext.EntityContext)),
 				SegmentID: int64(segment.ID),
 			}
-			return nil, log, false
+			return nil, log, true
```
zhouzhuojie: Caught by integration tests. We need to continue if `evalContext.EntityContext` if nil or invalid, because there may be some segments below that has no constraints.

PR #208, opened 2019-01-30
Add retry for DB connection

### LINKED ISSUE #207: Retry DB Connections
Related to #110

## Expected Behavior
When starting the application, if a DB connection can not be initiated, the application should retry before failing and terminating.

## Current Behavior
If the application fails to connect to a DB it immediately exits.

## Possible Solution
A simple solution could be to allow for a small number of connection attempts at a fixed interval before calling ```logrus.Fatal```. A more complicated solution could involve exponential backoff or and/or a configurable wait period before failure.

## Steps to Reproduce
Configure Flagr to use PostgreSQL or MYSQL database.
Start Flagr before starting database server.

## Context
We have Flagr running in Kubernetes on Google Cloud Platform. It is configured to use a PostgreSQL database. In order to access that database, we have a sidecar which runs Cloud SQL Proxy. When deploying the server, Flagr attempts to access the database while the sidecar container is still initializing. This forces Kubernetes to restart the Flagr container upon which it is able to connect to the database through the proxy.

While this approach technically works, it seems like Flagr should be able to tolerate a small delay that could occur when getting SQL servers running in a containerized environment.

zhouzhuojie:
## Description
Fix #207 

## Motivation and Context
Add retries so that the DB connection is more reliable when we start the flagr server.

## How Has This Been Tested?
Test locally with sqlite3. Will go through integration tests

## Types of changes

## Checklist:


pkg/entity/db.go
```diff
@@ -30,10 +31,22 @@ var AutoMigrateTables = []interface{}{
...
 }
 
+func connectDB() (db *gorm.DB, err error) {
+	err = retry.Do(
+		func() error {
+			db, err = gorm.Open(config.Config.DBDriver, config.Config.DBConnectionStr)
+			return err
+		},
+		retry.Attempts(config.Config.DBConnectionRetryAttempts),
+		retry.Delay(config.Config.DBConnectionRetryDelay),
```
zhouzhuojie: btw, the default retryType is backoff.

https://github.com/avast/retry-go/blob/master/retry.go#L80-L87

pkg/config/env.go
```diff
@@ -42,6 +42,9 @@ var Config = struct {
 	// DBConnectionDebug controls whether to show the database connection debugging logs
 	// warning: it may log the credentials to the stdout
 	DBConnectionDebug bool `env:"FLAGR_DB_DBCONNECTION_DEBUG" envDefault:"true"`
+	// DBConnectionRetryAttempts controls how we are going to retry on db connection when start the flagr server
```
zhouzhuojie: a note to myself, we probably need to document in the env.go, what are the minimal env variables do we need for a new setup.

pkg/entity/db.go
crberube: Awesome. By default 10 retries gives us 25.5 seconds of reconnect time. If we change the default attempts to 9, our retry time comes down to 12.7 seconds. That seems more like a happy medium between long enough and not too long to me. What do you think?

Issue #203, opened 2019-01-10
Consolidate the evalResult for kafka, kinesis and pubsub logging


zhouzhuojie:
It would be great to merge the evalResult struct for kafka and kinesis. It make future vendors for data recording easier.

https://github.com/checkr/flagr/blob/master/pkg/handler/data_recorder_kafka.go#L136
https://github.com/checkr/flagr/blob/master/pkg/handler/data_recorder_kinesis.go#L88

```
type kafkaMessageFrame struct {
	Payload   string `json:"payload"`
	Encrypted bool   `json:"encrypted"`
}
```

```
type kinesisMessageFrame struct {
	Payload   string `json:"payload"`
	Encrypted bool   `json:"encrypted"`
}
```

zhouzhuojie:
Note, we also need to consolidate pubsub as well

PR #209, opened 2019-01-30
Added Google Cloud Pubsub support for data records

vic3lord:
## Description

Adding Google Cloud Pubsub support for data records, currently just a minimal implementation

## Motivation and Context

We are using Google Cloud Pubsub and needed something native for data records

## How Has This Been Tested?

running `make` with all tasks, passed all tests and started flagr server

## Types of changes

## Checklist:


crberube:
Awesome. We are also looking to use Pub/Sub as our data pipeline.

pkg/handler/data_recorder_pubsub.go
```diff
@@ -0,0 +1,69 @@
...
+	producer *pubsub.Client
+	topic    *pubsub.Topic
+}
+
+// NewPubsubRecorder creates a new Pubsub recorder
+var NewPubsubRecorder = func() DataRecorder {
+	client, err := pubsub.NewClient(
+		context.Background(),
+		config.Config.RecorderPubsubProjectID,
+		option.WithServiceAccountFile(config.Config.RecorderPubsubKeyFile),
```
crberube: This method is deprecated, it is recommended to use ```WithCredentialsFile``` instead. Just starting to look into this to check their equivalency, but they seem to operate on the same idea.

As this is an option, does setting it to an empty string affect the use of Application Default Credentials? Would want to make sure that is not the case.

pkg/handler/data_recorder_pubsub.go
```diff
@@ -0,0 +1,69 @@
...
+	res := p.topic.Publish(ctx, &pubsub.Message{Data: payload})
+	if config.Config.RecorderPubsubVerbose {
+		id, err := res.Get(ctx)
+		if err != nil {
+			logrus.WithField("pubsub_error", err).Errorf("error pushing to pubsub: %v", id)
+		}
+	}
+}
+
+type pubsubEvalResult struct {
```
zhouzhuojie: related to https://github.com/checkr/flagr/issues/203, I think it would be great if pubsub also follow the same pattern. The better if we can fix #203 and have a single frame struct for data records.

pkg/handler/data_recorder_pubsub.go
vic3lord: I'm sorry, you are right! this option is deprecated I just used a chunk of code I am already using for the past 2 years in many go services here... I am pushing a fix now (thanks I'm also fixing my own services)

Regarding the empty string, yes it will work with default creds, matter effect this is exactly how we develop on our machines using our own accounts with proper IAM

pkg/handler/data_recorder_pubsub.go
vic3lord: I totally agree, I looked at the code and tried to follow your style just to make sure everything will pass and be ok on your end... But sure I think there's a lot of place to improve some of the things.

I will take a look at those relevant issues and will update ASAP (this is night over here)

pkg/handler/data_recorder_pubsub.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		logrus.WithField("pubsub_error", err).Error("error marshaling")
+	}
+
+	ctx := context.Background()
+	res := p.topic.Publish(ctx, &pubsub.Message{Data: payload})
+	if config.Config.RecorderPubsubVerbose {
+		id, err := res.Get(ctx)
+		if err != nil {
+			logrus.WithField("pubsub_error", err).Errorf("error pushing to pubsub: %v", id)
```
zhouzhuojie: id has high entropy, let's also move it into `WithFields`

pkg/handler/data_recorder_pubsub.go
zhouzhuojie: thanks! I think you can use 
```
type pubsubMessageFrame struct {
	Payload   string `json:"payload"`
	Encrypted bool   `json:"encrypted"`
}
```

Consolidating 2 structs is the same amout of work for 3. We can fix #203 later.

Also, make sure the final payload is a JSON struct looks like
```
{
  "payload": "<json marshal of EvalResult>",
  "encrypted": false
}
```

zhouzhuojie: and test coverage of this file :)

vic3lord: @zhouzhuojie regarding the pubsubMessageFrame, I added it on my end but it looks weird to marshal into json string because pubsub accepts []byte anyways so the conversion happens for no reason... I'm pushing the code to share this with you, LMK what you think

crberube:
Need to check further to see if this is a larger-scale thing, but it seems that if the pubsub connection fails then an EvaluationResult fails to be returned. @zhouzhuojie would this be expected behavior? It seems to me that we'd want to return a result anyways given that the data recording should be an async process.

pkg/handler/data_recorder_pubsub.go
zhouzhuojie: I knew the json string payload at the first look is weird. The reason I want to standardize the log format is that 

- Extensibility. If people want to end-to-end encryption, compression, or other meta things related to the message frame, they can do it regardless of the choice of the recorder type. They can add more fields into the message frame struct. Payload as a string fits into this decision.
- Portability. The data analytics pipeline requires no changes if one wants to switch between providers.

zhouzhuojie:
> Need to check further to see if this is a larger-scale thing, but it seems that if the pubsub connection fails then an EvaluationResult fails to be returned. @zhouzhuojie would this be expected behavior? It seems to me that we'd want to return a result anyways given that the data recording should be an async process.

Why do you think "if the pubsub connection fails then an EvaluationResult fails to be returned"? The connection won't affect online evaluation. AsyncRecord function should just log errors if there's any failure, and it shouldn't panic or exit as well.

pkg/handler/data_recorder_pubsub.go
```diff
@@ -0,0 +1,89 @@
...
+		Encrypted: false,
+	}
+
+	_, err = messageFrame.encode()
+	if err != nil {
+		logrus.WithField("pubsub_error", err).Error("error marshaling")
+	}
+
+	ctx := context.Background()
+	res := p.topic.Publish(ctx, &pubsub.Message{Data: payload})
```
zhouzhuojie: if you are using messageFrame, here the payload should be the return of `messageFrame.encode()`

crberube:
Here is what I'm seeing:

1) Start up Flagr (default SQLite, w/ data recording enabled. in this case pubsub). PubSub is not running to simulate connection issue.
2) Issue request:
```
curl --header "Content-Type: application/json" --request POST --data '{"flagId":1}' localhost:18000/api/v1/evaluation
curl: (52) Empty reply from server
```

Server logs:
```
INFO[0248] started handling request                      method=POST remote="172.17.0.1:34236" request=/api/v1/evaluation
{"FlagEvalResult":{"evalContext":{"entityID":"randomly_generated_544474078","flagID":1},"evalDebugLog":{"segmentDebugLogs":[]},"flagID":1,"flagKey":"kmmcd1nsd6ze56chh","flagSnapshotID":9,"segmentID":1,"timestamp":"2019-01-31T19:57:38Z","variantAttachment":null,"variantID":null,"variantKey":null}}
ERRO[0308] error pushing to pubsub                       id= pubsub_error="context deadline exceeded"
INFO[0308] completed handling request                    measure#flagr.latency=60002253200 method=POST remote="172.17.0.1:34236" request=/api/v1/evaluation status=200 text_status=OK took=1m0.0022532s
```

I'd expect the result to be returned immediately, but it's never returned at all. In the OpenAPI client for Go, this ends up setting the error value of the '''PostEvaluationResult''' method.

It looks like it's hanging on the ```pubsub.NewClient``` part of the code, but again, still digging.

zhouzhuojie:
@crberube @vic3lord 

I see. Inside `NewPubsubRecorder` should do `logrus.Fatal` instead of just `logrus.Error` when we get error from `pubsub.NewClient`.

pkg/handler/data_recorder_pubsub.go
```diff
@@ -0,0 +1,89 @@
...
+
+// NewPubsubRecorder creates a new Pubsub recorder
+var NewPubsubRecorder = func() DataRecorder {
+	client, err := pubsub.NewClient(
+		context.Background(),
+		config.Config.RecorderPubsubProjectID,
+		option.WithCredentialsFile(config.Config.RecorderPubsubKeyFile),
+	)
+	if err != nil {
+		logrus.WithField("pubsub_error", err).Error("error getting pubsub client")
```
zhouzhuojie: let's `Fatal` here instead of `Error`

tosonggao: I think the issue is caused by `get` method from pubsub which is a blocking operation, see pubsub doc [here](https://godoc.org/cloud.google.com/go/pubsub#PublishResult.Get)
we should pass context either with a deadline or timeout to `get` method  instead of just background context.

crberube: Can confirm this is the issue. When I disable verbose logging, everything works as expected.

pkg/handler/data_recorder_pubsub.go
vic3lord: That is what I asked earlier, Data is a `[]byte` but encode returns a formatted json string I can cast it into `[]byte` but then what's the point?

zhouzhuojie: The point is the metadata fields (e.g. encrypted, compressed, future additional fields) of the frame should also be included into the Data field.

If we don't put it now, future changes are all breaking changes.

pkg/handler/data_recorder_pubsub.go
```diff
@@ -61,9 +61,11 @@ func (p *pubsubRecorder) AsyncRecord(r *models.EvalResult) {
 		logrus.WithField("pubsub_error", err).Error("error marshaling")
 	}
 
-	ctx := context.WithTimeout(context.Background(), 10*time.Second) // TODO: make it configurable
+	ctx := context.Background()
 	res := p.topic.Publish(ctx, &pubsub.Message{Data: payload})
 	if config.Config.RecorderPubsubVerbose {
+		ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
```
crberube: What do you think about combining this with running the Get within a new goroutine?

zhouzhuojie: +1 for another goroutine. Also, 5s can be moved into env config.

vic3lord: Agreed, I added both

vic3lord:
I need some help regarding the tests, I added test coverage to pubsub but now the `data_recorder_test.go` fails, I don't really understand how it passed before because it used the same `NewPubsubRecorder()` function when calling `GetDataRecorder()`.

I can see that we use in all data recorders with their production functions without mocking, meaning connection should fail on all of them, unless I am missing something that runs as a mock server on CI

crberube:
Still figuring out the best way to handle this, but I know what is wrong currently:

When the following code is called:
```go
	client, err := pubsub.NewClient(
		context.Background(),
		config.Config.RecorderPubsubProjectID,
		option.WithCredentialsFile(config.Config.RecorderPubsubKeyFile),
	)
```

the library uses the following rules to create a new client:

> GCP client libraries use a strategy called Application Default Credentials (ADC) to find your application's credentials. When your code uses a client library, the strategy checks for your credentials in the following order:

> 1. First, ADC checks to see if the environment variable GOOGLE_APPLICATION_CREDENTIALS is set. If the variable is set, ADC uses the service account file that the variable points to. The next section describes how to set the environment variable.

> 2. If the environment variable isn't set, ADC uses the default service account that Compute Engine, Kubernetes Engine, App Engine, and Cloud Functions provide, for applications that run on those services.

> 3. If ADC can't use either of the above credentials, an error occurs.

On our machines, step 2 is happening, likely because we have run a `gcloud auth` command in the past. On the CircleCI containers, this has never been run and so we go to step 3 (error).

NewClient signature is as follows:

```go
func NewClient(ctx context.Context, projectID string, opts ...option.ClientOption) (c *Client, err error) {
```

and the return value in case of an error is this:
```go
return nil, fmt.Errorf("pubsub: %v", err)
```

so when the rest of the code executes:
```go
	if err != nil {
		// TODO: use Fatal again after fixing the test expecting to not panic.
		// logrus.WithField("pubsub_error", err).Fatal("error getting pubsub client")
		logrus.WithField("pubsub_error", err).Error("error getting pubsub client")
	}

	return &pubsubRecorder{
		producer: client,
		topic:    client.Topic(config.Config.RecorderPubsubTopicName),
		enabled:  config.Config.RecorderEnabled,
	}
```

client is a null pointer, and you get a null pointer dereference.

So that's the issue.

Having the Fatal call will prevent us from dereferencing a nil pointer which is good. It is also the same way we handle error cases in the kafka client so it would follow the standards there.

So to fix the issue in the test we need to either
a) allow the client to retrieve credentials or
b) mock the client somehow

pkg/config/env.go
```diff
@@ -97,6 +97,13 @@ var Config = struct {
 	RecorderKinesisAggregateBatchSize  int           `env:"FLAGR_RECORDER_KINESIS_AGGREGATE_BATCH_SIZE" envDefault:"51200"`
 	RecorderKinesisVerbose             bool          `env:"FLAGR_RECORDER_KINESIS_VERBOSE" envDefault:"false"`
 
+	// Pubsub related configurations for data records logging (Flagr Metrics)
+	RecorderPubsubProjectID     string        `env:"FLAGR_RECORDER_PUBSUB_PROJECT_ID" envDefault:""`
+	RecorderPubsubTopicName     string        `env:"FLAGR_RECORDER_PUBSUB_TOPIC_NAME" envDefault:"flagr-records"`
+	RecorderPubsubKeyFile       string        `env:"FLAGR_RECORDER_PUBSUB_KEYFILE" envDefault:""`
+	RecorderPubsubVerbose       bool          `env:"FLAGR_RECORDER_PUBSUB_VERBOSE" envDefault:"false"`
+	RecorderPubsubVerboseCancel time.Duration `env:"FLAGR_RECORDER_PUBSUB_VERBOSE_CANCEL" envDefault:"5s"`
```
zhouzhuojie: how about `RecorderPubsubVerboseCancelTimeout`? Otherwise, it's not clear to me that `cancel` can be a time.Duration.

crberube:
@vic3lord I've created a PR on your fork which fixes the above issues

pkg/handler/data_recorder_pubsub.go
```diff
@@ -0,0 +1,99 @@
...
+		return
+	}
+
+	pr := &pubsubEvalResult{
+		EvalResult: r,
+	}
+
+	payload, err := pr.Payload()
+	if err != nil {
+		logrus.WithField("pubsub_error", err).Error("error marshaling")
```
crberube: Two things:

1) It would be good if these errors (when generating payload and when encoding the message frame) were distinct from one another. Maybe change the error message on this one to "error marshaling payload" and "error marshaling message frame" on the other? Or something to that effect.

2) In both these cases (```encode()``` and ```Payload()```), during the marshaling step the returned values are ```([]byte, error)```. If an error is returned, that means that the payload is an empty byte slice. So in effect what will happen is that a blank message will be written to pubsub. 

In the rare even that this occurs, would it make more sense to return after logging the error? So in each case, have something like:
```go
if err != nil {
  logrus.WithField...
  return
}
```

crberube:
This looks good to me pending my question about handling potential marshal errors.

pkg/handler/data_recorder_pubsub.go
vic3lord: Yep, I do the same on our services... I am trying to comply with your style guide

crberube:
Cool. @zhouzhuojie how are you feeling about everything at this point?

pkg/handler/data_recorder_pubsub.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"context"
+	"encoding/json"
+
+	"cloud.google.com/go/pubsub"
+	"github.com/checkr/flagr/pkg/config"
+	"github.com/checkr/flagr/swagger_gen/models"
+	"google.golang.org/api/option"
+
```
zhouzhuojie: nit, no need a new line here

pkg/handler/data_recorder_pubsub.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+type pubsubRecorder struct {
+	enabled  bool
```
zhouzhuojie: is this `enabled` necessary here? eval.go checks it here https://github.com/checkr/flagr/blob/master/pkg/handler/eval.go#L178

crberube: oh yeah... this could probably be removed from the kafka and kinesis versions as they do the check as well

zhouzhuojie: didn't realize that it was also used in kafka and kinesis, we can probably clean them up in other PRs

vic3lord:
Thank you so much for helping, I fixed the little nits.

Where would it be best to document the pubsub addition and its auth docs?

zhouzhuojie:
> Thank you so much for helping, I fixed the little nits.
> 
> Where would it be best to document the pubsub addition and its auth docs?

I think you can put it after the Kinesis section in https://github.com/checkr/flagr/blob/master/docs/flagr_env.md

Issue #196, opened 2018-12-13
Support Prometheus


zhouzhuojie:
Support Prometheus for better observability

jasongwartz:
@zhouzhuojie We're starting to test out flagr at Ecosia, and we use prometheus monitoring heavily, so I'd be interested in taking this issue if it still needs doing. Is it just a matter of exporting metrics for the http handlers?

zhouzhuojie:
Yes, metrics for the HTTP handlers could be the first step! Also, for the evaluation result, I think we can export metrics just like what we are doing for datadog.
https://github.com/checkr/flagr/blob/master/pkg/handler/eval.go#L185-L200

PR #221, opened 2019-02-19
Adds Prometheus export (requests and variant eval)

jasongwartz:
## Description
This PR adds a Prometheus exporter based on the prometheus client-go package. It exports metrics regarding the requests (a count of requests as well as optionally a histogram of request latencies), and a count of variant allocations per flag (similar to datadog statsd).

## Motivation and Context

Closes https://github.com/checkr/flagr/issues/196
We are trialing flagr at Ecosia, and we use Prometheus for monitoring all our systems running Kubernetes.

## How Has This Been Tested?
Added unit tests for config, including creating the prometheus metrics (which will panic if they cannot be created). I also built the docker image and am running it in our dev cluster.

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly. (It seems like the embedded env.go will automatically be updated after merging)


pkg/handler/eval.go
```diff
@@ -199,6 +200,19 @@ var logEvalResultToDatadog = func(r *models.EvalResult) {
 	)
 }
 
+var logEvalResultToPrometheus = func(r *models.EvalResult) {
+	if config.Global.Prometheus.EvalCounter == nil {
+		return
+	}
+	config.Global.Prometheus.EvalCounter.WithLabelValues(
+		util.SafeStringWithDefault(r.EvalContext.EntityType, "null"),
```
zhouzhuojie: not very familiar with the convention of prometheus lable values, does it need to have a name in the value? For example,

```
fmt.Sprintf("flag_id:%s", util.SafeStringWithDefault(r.FlagID, "null"))
```

Otherwise, how can we know if it's variantID or flagID, or something else?

jasongwartz: @zhouzhuojie that's a bit confusing, yeah - the labels are coded in when the Counter is created here: https://github.com/checkr/flagr/pull/221/files#diff-091ace02b49a46f55f2586cfec89a160R96

pkg/config/config_test.go
```diff
@@ -29,3 +30,25 @@ func TestSetupStatsd(t *testing.T) {
...
+	Config.PrometheusEnabled = true
+	setupPrometheus()
+	assert.NotNil(t, Global.Prometheus.EvalCounter)
+	assert.NotNil(t, Global.Prometheus.RequestCounter)
+	assert.Nil(t, Global.Prometheus.RequestHistogram)
+}
+
+func TestSetupPrometheusWithLatencies(t *testing.T) {
+	prometheus.DefaultRegisterer = prometheus.NewRegistry()
+	Config.PrometheusEnabled = true
```
zhouzhuojie: Can you reset the configuration after the test?

pkg/config/config_test.go
```diff
@@ -29,3 +30,25 @@ func TestSetupStatsd(t *testing.T) {
 	})
 	Config.StatsdEnabled = false
 }
+
+func TestSetupPrometheus(t *testing.T) {
+	prometheus.DefaultRegisterer = prometheus.NewRegistry()
+	Config.PrometheusEnabled = false
+	setupPrometheus()
+	assert.Nil(t, Global.Prometheus.EvalCounter)
+	Config.PrometheusEnabled = true
```
zhouzhuojie: same here

pkg/config/config_test.go
jasongwartz: sure, no problem

pkg/config/config_test.go
jasongwartz: done, just pushed

pkg/config/env.go
```diff
@@ -67,6 +67,13 @@ var Config = struct {
 	StatsdAPMPort        string `env:"FLAGR_STATSD_APM_PORT" envDefault:"8126"`
 	StatsdAPMServiceName string `env:"FLAGR_STATSD_APM_SERVICE_NAME" envDefault:"flagr"`
 
+	// PrometheusEnabled - enable prometheus metrics export
+	PrometheusEnabled bool `env:"FLAGR_PROMETHEUS_ENABLED" envDefault:"false"`
+	// PrometheusPath - set the path on which prometheus metrics are available to scrape
+	PrometheusPath string `env:"FLAGR_PROMETHEUS_PATH" envDefault:"/metrics"`
+	// PrometheusIncludeLatencyHistogram - set whether Prometheus should also export a histogram of request latencies (this increases cardinality significantly)
```
zhouzhuojie: +1, good to know

jasongwartz:
@zhouzhuojie thanks!

PR #233, opened 2019-03-08
Refactor data recorder

### LINKED ISSUE #232: Data Pipeline Format
Right now there are a couple of quirks with the way that result data in formatted through the data pipeline. The biggest issue for us right now is the use of `MarshalIdent` on the JSON. I see that it was used in the past for some sort of Kafka Connector, but the result leaves us with JSON where each element begins on a new line with a prefix. This breaks a lot of data processing applications that expect JSON elements to be newline delimited.

A second quirk I noticed is that the JSON message frame saves the payload as a string of escaped JSON. I suspect this has something to do with the encryption option for Kafka? However, when things are not encrypted, it would be much cleaner if the all of the JSON was marshaled at once such that the payload field was simply another level of the overall JSON object as opposed to a string containing the JSON data.

I realize that these changes would amount to breaking changes in the data generated by the data pipeline, but would it be reasonable to add some sort of configuration option(s) to allow for these changes? This is something that I could address ASAP if we can figure out a way to cleanly handle this.

zhouzhuojie:
## Description
Fix #232 #203 

- Refactor all data recorders
- Add data_record_frame.go
- Change the interface of DataRecorder to comply with DataRecordFrame

## Motivation and Context
- Add the option to set the output mode of DataRecordFrame
- Consolidate and simplify adding new data recorders

## How Has This Been Tested?
Tested locally, note that `kt` tool prints the value field of kafka messages with escape.

```
FLAGR_RECORDER_ENABLED=true FLAGR_RECORDER_FRAME_OUTPUT_MODE=payload_raw_json make run

kt consume -topic flagr-records -offsets all=newest-1:
{
  "partition": 0,
  "offset": 8,
  "key": "a1234",
  "value": "{\"payload\":{\"evalContext\":{\"enableDebug\":true,\"entityContext\":{\"hello\":\"world\"},\"entityID\":\"a1234\",\"entityType\":\"a123\",\"flagID\":1,\"flagKey\":\"kmmcd1nsd6\"},\"evalDebugLog\":{\"segmentDebugLogs\":[{\"msg\":\"argument: state not found\",\"segmentID\":1}]},\"flagID\":1,\"flagKey\":\"kmmcd1nsd6\",\"flagSnapshotID\":46,\"segmentID\":1,\"timestamp\":\"2019-03-08T23:45:37Z\",\"variantAttachment\":null,\"variantID\":null,\"variantKey\":null}}",
  "timestamp": "1969-12-31T15:59:59.999-08:00"
}


FLAGR_RECORDER_ENABLED=true make run

kt consume -topic flagr-records -offsets all=newest-1:
{
  "partition": 0,
  "offset": 9,
  "key": "a1234",
  "value": "{\"payload\":\"{\\\"evalContext\\\":{\\\"enableDebug\\\":true,\\\"entityContext\\\":{\\\"hello\\\":\\\"world\\\"},\\\"entityID\\\":\\\"a1234\\\",\\\"entityType\\\":\\\"a123\\\",\\\"flagID\\\":1,\\\"flagKey\\\":\\\"kmmcd1nsd6\\\"},\\\"evalDebugLog\\\":{\\\"segmentDebugLogs\\\":[{\\\"msg\\\":\\\"argument: state not found\\\",\\\"segmentID\\\":1}]},\\\"flagID\\\":1,\\\"flagKey\\\":\\\"kmmcd1nsd6\\\",\\\"flagSnapshotID\\\":46,\\\"segmentID\\\":1,\\\"timestamp\\\":\\\"2019-03-08T23:47:23Z\\\",\\\"variantAttachment\\\":null,\\\"variantID\\\":null,\\\"variantKey\\\":null}\",\"encrypted\":false}",
  "timestamp": "1969-12-31T15:59:59.999-08:00"
}



```

## Types of changes

## Checklist:

### DIFF
--- a/pkg/config/env.go
+++ b/pkg/config/env.go
@@ -79,6 +79,23 @@ var Config = struct {
 	// RecorderType - the pipeline to log data records, e.g. Kafka
 	RecorderType string `env:"FLAGR_RECORDER_TYPE" envDefault:"kafka"`
 
+	/**
+	RecorderFrameOutputMode - indicates which data record frame output mode should we use.
+	Possible values: payload_string, payload_raw_json
+
+	* payload_string mode:
+		it respects the encryption settings, and it will stringify the payload to unify
+		the type of the output for both plaintext and encrypted payload.
+
+		{"payload":"{\"evalContext\":{\"entityID\":\"123\"},\"flagID\":1,\"flagKey\":null,\"flagSnapshotID\":1,\"segmentID\":1,\"timestamp\":null,\"variantAttachment\":null,\"variantID\":1,\"variantKey\":\"control\"}","encrypted": false}
+
+	* payload_raw_json mode:
+		it ignores the encryption settings.
+
+		{"payload":{"evalContext":{"entityID":"123"},"flagID":1,"flagKey":null,"flagSnapshotID":1,"segmentID":1,"timestamp":null,"variantAttachment":null,"variantID":1,"variantKey":"control"}}
+	*/
+	RecorderFrameOutputMode string `env:"FLAGR_RECORDER_FRAME_OUTPUT_MODE" envDefault:"payload_string"`
+
 	// Kafka related configurations for data records logging (Flagr Metrics)
 	RecorderKafkaVersion        string        `env:"FLAGR_RECORDER_KAFKA_VERSION" envDefault:"0.8.2.0"`
 	RecorderKafkaBrokers        string        `env:"FLAGR_RECORDER_KAFKA_BROKERS" envDefault:":9092"`
--- a/pkg/handler/data_record_frame.go
+++ b/pkg/handler/data_record_frame.go
@@ -1,0 +1,98 @@
+package handler
+
+import (
+	"encoding/base64"
+	"encoding/json"
+
+	"github.com/brandur/simplebox"
+	"github.com/checkr/flagr/pkg/util"
+	"github.com/checkr/flagr/swagger_gen/models"
+)
+
+type dataRecordEncryptor interface {
+	Encrypt([]byte) (string, error)
+}
+
+type simpleboxEncryptor struct{ key [simplebox.KeySize]byte }
+
+func (se *simpleboxEncryptor) Encrypt(b []byte) (string, error) {
+	s := base64.StdEncoding.EncodeToString(
+		simplebox.NewFromSecretKey(&se.key).Encrypt(b),
+	)
+	return s, nil
+}
+
+func newSimpleboxEncryptor(k string) dataRecordEncryptor {
+	key := [simplebox.KeySize]byte{}
+	copy(key[:], k)
+	return &simpleboxEncryptor{key: key}
+}
+
+const (
+	frameOutputModePayloadRawJSON = "payload_raw_json"
+)
+
+// DataRecordFrameOptions represents the options we can set to create a DataRecordFrame
+type DataRecordFrameOptions struct {
+	Encrypted       bool
+	Encryptor       dataRecordEncryptor
+	FrameOutputMode string
+}
+
+type rawPayload struct {
+	Payload json.RawMessage `json:"payload"`
+}
+
+type stringPayload struct {
+	Payload   string `json:"payload"`
+	Encrypted bool   `json:"encrypted"`
+}
+
+// DataRecordFrame represents the structure we can json.Marshal into data recorders
+type DataRecordFrame struct {
+	evalResult models.EvalResult
+	options    DataRecordFrameOptions
+}
+
+// MarshalJSON defines the behavior of MarshalJSON for DataRecordFrame
+func (drf *DataRecordFrame) MarshalJSON() ([]byte, error) {
+	payload, err := drf.evalResult.MarshalBinary()
+	if err != nil {
+		return nil, err
+	}
+
+	if drf.options.FrameOutputMode == frameOutputModePayloadRawJSON {
+		return json.Marshal(&rawPayload{
+			Payload: payload,
+		})
+	}
+
+	if drf.options.Encrypted && drf.options.Encryptor != nil {
+		encryptedPayload, err := drf.options.Encryptor.Encrypt(payload)
+		if err != nil {
+			return nil, err
+		}
+		return json.Marshal(&stringPayload{
+			Payload:   encryptedPayload,
+			Encrypted: true,
+		})
+	}
+
+	return json.Marshal(&stringPayload{
+		Payload:   string(payload),
+		Encrypted: false,
+	})
+}
+
+// GetPartitionKey gets the partition key from entityID
+func (drf *DataRecordFrame) GetPartitionKey() string {
+	if drf.evalResult.EvalContext == nil {
+		return ""
+	}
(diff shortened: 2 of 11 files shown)


pkg/config/env.go
```diff
@@ -79,6 +79,21 @@ var Config = struct {
 	// RecorderType - the pipeline to log data records, e.g. Kafka
 	RecorderType string `env:"FLAGR_RECORDER_TYPE" envDefault:"kafka"`
 
+	// RecorderFrameOutputMode - indicates which data record frame output mode should we use.
+	// Possible values: payload, payload_raw_json
+	//
+	// payload mode - it respects the encryption settings, and it will stringify the payload to unify
+	//                the type of the output for both plaintext and encrypted payload.
+	//   {
```
zhouzhuojie: TODO: going to add concrete examples here

zhouzhuojie:
TODO, increase test coverage for `date_record_frame.go`

pkg/handler/data_record_frame.go
```diff
@@ -0,0 +1,92 @@
...
+type DataRecordFrameOptions struct {
+	Encrypted       bool
+	Encryptor       dataRecordEncryptor
+	FrameOutputMode string
+}
+
+// DataRecordFrame represents the structure we can json.Marshal into data recorders
+type DataRecordFrame struct {
+	Encrypted      bool            `json:"encrypted"`
+	Payload        string          `json:"payload"`
```
crberube: I think that at the very least we should add `omitempty` to both `Payload` and `PayloadRawJSON` as there is only ever going to be one or the other. The more I've though about this however I've come to think that it would be nice to achieve a couple of different goals:

1) Keep the output JSON message as `payload` whether or not options are set to use raw payload or not. In each case, the resulting data is the payload.
2) Forgo the "encrypted" field if we have the raw JSON setting turned on. If we are outputting raw JSON then the result will by definition not be encrypted.

With these ideas in mind, I've created an alternative version of the this file that takes these scenarios into account. I'm sure there is room for some cleanup but I'd like to see what you think of this idea:

```go
package handler

import (
	"encoding/base64"
	"encoding/json"

	"github.com/brandur/simplebox"
	"github.com/checkr/flagr/pkg/util"
	"github.com/checkr/flagr/swagger_gen/models"
)

type dataRecordEncryptor interface {
	Encrypt([]byte) (string, error)
}

type simpleboxEncryptor struct{ key [simplebox.KeySize]byte }

func (se *simpleboxEncryptor) Encrypt(b []byte) (string, error) {
	s := base64.StdEncoding.EncodeToString(
		simplebox.NewFromSecretKey(&se.key).Encrypt(b),
	)
	return s, nil
}

func newSimpleboxEncryptor(k string) dataRecordEncryptor {
	key := [simplebox.KeySize]byte{}
	copy(key[:], k)
	return &simpleboxEncryptor{key: key}
}

const (
	frameOutputModePayloadRawJSON = "payload_raw_json"
)

// DataRecordFrameOptions represents the options we can set to create a DataRecordFrame
type DataRecordFrameOptions struct {
	Encrypted       bool
	Encryptor       dataRecordEncryptor
	FrameOutputMode string
}

type rawPayload struct {
	Payload json.RawMessage `json:"payload"`
}

type stringPayload struct {
	Payload   string `json:"payload"`
	Encrypted bool
}

// DataRecordFrame represents the structure we can json.Marshal into data recorders
type DataRecordFrame struct {
	evalResult models.EvalResult
	options    DataRecordFrameOptions
}

func (drf *DataRecordFrame) MarshalJSON() ([]byte, error) {
	payload, err := drf.evalResult.MarshalBinary()
	if err != nil {
		return nil, err
	}

	if drf.options.FrameOutputMode == frameOutputModePayloadRawJSON {
		return json.Marshal(&rawPayload{
			Payload: payload,
		})
	}

	if drf.options.Encrypted && drf.options.Encryptor != nil {
		encryptedPayload, err := drf.options.Encryptor.Encrypt(payload)
		if err != nil {
			return nil, err
		}
		return json.Marshal(&stringPayload{
			Payload:   encryptedPayload,
			Encrypted: true,
		})
	}

	return json.Marshal(&stringPayload{
		Payload:   string(payload),
		Encrypted: false,
	})
}

// GetPartitionKey gets the partition key from entityID
func (drf *DataRecordFrame) GetPartitionKey() string {
	if drf.evalResult.EvalContext == nil {
		return ""
	}
	return util.SafeString(drf.evalResult.EvalContext.EntityID)
}

// Output sets the paylaod using its input and returns the json marshal bytes
func (drf *DataRecordFrame) Output() ([]byte, error) {
	return json.Marshal(drf)
}

```

zhouzhuojie: I like it, much better!

pkg/config/env.go
```diff
@@ -79,6 +79,34 @@ var Config = struct {
...
+	// Possible values: payload_string, payload_raw_json
+	//
+	// payload_string mode - it respects the encryption settings, and it will stringify the payload to unify
+	//                the type of the output for both plaintext and encrypted payload.
+	//   {
+	//     "payload": "{\"evalContext\":{\"entityID\":\"123\"},\"flagID\":1,\"flagKey\":null,\"flagSnapshotID\":1,\"segmentID\":1,\"timestamp\":null,\"variantAttachment\":null,\"variantID\":1,\"variantKey\":\"control\"}",
+	//     "encrypted": false
+	//   }
+	//
+	// payload_raw_json mode - ignores the encryption settings.
```
crberube: Since we have removed the `MarshalIdent` I think we should either put the example JSON in a single line (as in the string version) or make a note in the comment that the resulting output will consist of a single record per line.

zhouzhuojie: good catch, I changed to show the single-line version in the comments.

pkg/handler/data_recorder.go
```diff
@@ -14,7 +14,8 @@ var (
 
 // DataRecorder can record and produce the evaluation result
 type DataRecorder interface {
-	AsyncRecord(*models.EvalResult)
+	AsyncRecord(models.EvalResult)
```
crberube: Seems like there should be some standardization on whether EvalResult is used as a pointer or a value. It seems to be a bit of both right now. Any thoughts on this?

crberube:
Overall this looks pretty solid. I think at this point my only question has to do with the way EvalResult is treated (as noted above). Thanks for the help with this!

pkg/handler/data_recorder.go
zhouzhuojie: When I refactored this interface, I had to make too many nil checks all over the places, and they didn't give much margin benefits especially when the struct is small enough. Instead, I will just assume that we can receive a struct from upstream. Which kind of makes sense: if there's nothing to log, it shouldn't even bother any DataRecorders.

And in the evaluation code path, the ultimate response is auto generated from https://github.com/checkr/flagr/blob/master/swagger_gen/restapi/operations/evaluation/post_evaluation_responses.go#L28, which is a pointer. Therefore the pointer reference is prepared inside eval.go

crberube: I see, that makes sense. Well that's all I really have from my end. Looks good.

PR #235, opened 2019-03-09
Add EvalDebugEnabled env variable

### LINKED ISSUE #228: Disable enableDebug Propery for Evaluation Transactions
There does not seem to be a way to disable the enableDebug option for Evaluations in a Production environment. It seems like that could lead to some data about the A/B Tests being run to leak to external users.
Is there any plan to create a config to disable that?

zhouzhuojie:
## Description
Fix #228

## Motivation and Context
Add an option so that we can disable evaluation debug globally due to security reasons. By default, it's enabled.

## Types of changes

## Checklist:

### DIFF
--- a/pkg/config/env.go
+++ b/pkg/config/env.go
@@ -30,6 +30,11 @@ var Config = struct {
 	EvalCacheRefreshTimeout time.Duration `env:"FLAGR_EVALCACHE_REFRESHTIMEOUT" envDefault:"59s"`
 	// EvalCacheRefreshInterval - time interval of getting the flags data from DB into the in-memory evaluation cache
 	EvalCacheRefreshInterval time.Duration `env:"FLAGR_EVALCACHE_REFRESHINTERVAL" envDefault:"3s"`
+	// EvalEnableDebug - controls if we want to return evaluation debugging information back to the api requests
+	// Note that this is a global switch:
+	//     if it's disabled, no evaluation debug info will be returned.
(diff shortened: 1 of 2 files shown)


pkg/config/env.go
```diff
@@ -30,6 +30,11 @@ var Config = struct {
 	EvalCacheRefreshTimeout time.Duration `env:"FLAGR_EVALCACHE_REFRESHTIMEOUT" envDefault:"59s"`
 	// EvalCacheRefreshInterval - time interval of getting the flags data from DB into the in-memory evaluation cache
 	EvalCacheRefreshInterval time.Duration `env:"FLAGR_EVALCACHE_REFRESHINTERVAL" envDefault:"3s"`
+	// EvalEnableDebug - controls if we want to return evaluation debugging information back to the api requests
+	// Note that this is a global switch:
+	//     if it's disabled, no evaluation debug info will be returned.
+	//     if it's enabled, it respects evaluation request's enableDebug field
+	EvalDebugEnabled bool `env:"FLAGR_EVAL_DEBUG_ENABLED" envDefault:"true"`
```
marceloboeira: True as default? I would make it less intrusive ...

zhouzhuojie: True is the current behavior. It controls globally if the server will need to respect the "enableDebug" filed from the request. https://checkr.github.io/flagr/api_docs/#operation/postEvaluation

marceloboeira: I wasn't aware, good to know 😄

Issue #216, opened 2019-02-06
[feat] block deletion if there are recent realtime traffic hitting the flag


zhouzhuojie:
For safety checks, we may want to block deletion if there are currently ongoing traffic hitting the flag. I can imagine there are multiple ways of doing so, but storing the real-time information across multiple flagr instances requires some tuning.

Possible solutions
- Introduce a shared cache for real-time information. E.g. Redis
- Periodically flush in-memory traffic data into the database, and then periodically load it from database and merge it with the current in-memory data.

PR #236, opened 2019-03-11
Add FlagRealtime entity

zhouzhuojie:
## Description
Prepare a realtime table to track the last evaluated at timestamp for each flag. This helps to prevent premature flag deletion, and also we can surface some realtime information to the API and UI. The goal is to minimize the evaluation API overhead, and at the safe time, achieve the eventual consistency via its own DB.

- Async write to in-memory cache
- Periodically dump to DB (if it has newer evaluated at timestamp)
- Future read will just directly from DB

## How Has This Been Tested?
Tested locally. TODO, more test coverage for `flag_realtime.go`

## Types of changes

## Checklist:

### DIFF
--- a/pkg/config/env.go
+++ b/pkg/config/env.go
@@ -30,6 +30,9 @@ var Config = struct {
 	EvalCacheRefreshTimeout time.Duration `env:"FLAGR_EVALCACHE_REFRESHTIMEOUT" envDefault:"59s"`
 	// EvalCacheRefreshInterval - time interval of getting the flags data from DB into the in-memory evaluation cache
 	EvalCacheRefreshInterval time.Duration `env:"FLAGR_EVALCACHE_REFRESHINTERVAL" envDefault:"3s"`
+	// EvalFlagRealtimeSyncInterval - time interval of saving to flag_realtimes table
+	// So worst case, DB will have the delay of EvalFlagRealtimeSyncInterval
+	EvalFlagRealtimeSyncInterval time.Duration `env:"FLAGR_EVAL_FLAG_REALTIME_SYNC_INTERVAL" envDefault:"30s"`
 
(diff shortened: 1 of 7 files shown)


zhouzhuojie:
Related to #216

pacoguzman:
Hey we're evaluating this tool and it looks pretty good so far. I'm interested to know how multiple instances deal with updates, as evaluation logic uses in memory data, how other instance know it's time to get info from the database.

I guess this PR will solve this issue to, is that right? Probably those instances doesn't need to be up to date at millisecond rates but I would like to know your thoughts about it.

Thanks in advance

zhouzhuojie:
> Hey we're evaluating this tool and it looks pretty good so far. I'm interested to know how multiple instances deal with updates, as evaluation logic uses in memory data, how other instance know it's time to get info from the database.
> 
> I guess this PR will solve this issue to, is that right? Probably those instances doesn't need to be up to date at millisecond rates but I would like to know your thoughts about it.
> 
> Thanks in advance

If you are referring to flags getting updated in the memory cache, you don't need to worry about that. Every flagr instance periodically pulls flags from either DB, local file, or remote HTTP endpoint. It's not at millisecond rates, it's controlled by `FLAGR_EVALCACHE_REFRESHINTERVAL`.

But for this PR, I'm still thinking about what works best. Since from https://github.com/checkr/flagr/pull/247, I decoupled DB from its evaluation cache source. The evaluation node may not even have a DB connection, which makes it hard to share real-time information across the nodes via DB.

pacoguzman:
Cool thanks for answering. We're successfully running flagr on production working great so far!

We're interesting on this PR too

> This helps to prevent premature flag deletion.

Do you need help to move this forward?

PR #238, opened 2019-03-19
Don't require fields on the evalResult response object

erithmetic:
This caused an issue under the following circumstance:
* A flag did not exist in flagr
* The consumer requested a flag by only specifying `:flag_key`
* Flagr returned a 200 response with `:flag_id` omitted
* The ruby code generated by swagger raised an `ArgumentError` because the response from the server was missing `:flag_id`

Ultimately, it's up to the consumer app anyway to decide what it requires from the server response, not the HTTP interface library.

### DIFF
--- a/docs/api_docs/bundle.yaml
+++ b/docs/api_docs/bundle.yaml
@@ -1105,15 +1105,6 @@ definitions:
         type: string
   evalResult:
     type: object
-    required:
-      - flagID
-      - flagKey
-      - segmentID
-      - variantID
-      - variantKey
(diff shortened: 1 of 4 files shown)


zhouzhuojie:
It requires changes more than just swagger yaml changes. Can you `make gen` and then help fix all the compilation error related to the changes?

zhouzhuojie:
It may be tricky to fix some tests, thus I created a new PR here https://github.com/checkr/flagr/pull/239

PR #239, opened 2019-03-19
Remove required fields from evalResult swagger model

zhouzhuojie:
## Description
This caused an issue under the following circumstance:
* A flag did not exist in flagr
* The consumer requested a flag by only specifying `:flag_key`
* Flagr returned a 200 response with `:flag_id` omitted
* The ruby code generated by swagger raised an `ArgumentError` because the response from the server was missing `:flag_id`

Ultimately, it's up to the consumer app anyway to decide what it requires from the server response, not the HTTP interface library.

## How Has This Been Tested?
Unit test and integration tests

## Types of changes

## Checklist:

### DIFF
--- a/pkg/config/middleware_test.go
+++ b/pkg/config/middleware_test.go
@@ -5,11 +5,8 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
-	"reflect"
 	"testing"
 
-	"github.com/DataDog/datadog-go/statsd"
-	"github.com/bouk/monkey"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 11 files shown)


pkg/config/middleware_test.go
```diff
@@ -293,41 +290,3 @@ func TestAuthMiddlewareWithUnauthorized(t *testing.T) {
 		}
 	})
 }
-
```
zhouzhuojie: Since go 1.12, `monkey.PatchInstanceMethod` is not working anymore, thus remove this hacky monkey patch

zhouzhuojie:
Tested the change with `rbflagr 0.1.2` (which still keeps the required fields of evalResult), looks good for the SDKs

PR #249, opened 2019-04-19
Fix lint and make vendor

zhouzhuojie:


pkg/handler/data_recorder_kinesis.go
```diff
@@ -21,7 +21,12 @@ type kinesisRecorder struct {
 
 // NewKinesisRecorder creates a new Kinesis recorder
 var NewKinesisRecorder = func() DataRecorder {
-	client := kinesis.New(session.New(aws.NewConfig()))
+	se, err := session.NewSession(aws.NewConfig())
```
zhouzhuojie: static check from golangci-lint caught this, `session.New` is deprecated. We should use `session.NewSession`

PR #258, opened 2019-05-21
Changed input text to json-editor in flag variant form

yosyad:
## Description
In order to edit easily the variant attachment, I replaced the existing input to the same json editor that used in the evaluator on the same page.

## Motivation and Context
I was trying to use flagr with a variant that has more than one key and found myself struggling with the input area. The existing input area was designed for using simple texts in forms, such as name or email, but not JSON.
In order to edit JSON, with syntax highlighting and auto-correct, out of the box, I found my self use the same component that was used on the same page.

## How Has This Been Tested?
Tested locally, manually.

## Types of changes

## Checklist:

### DIFF
--- a/browser/flagr-ui/src/components/DebugConsole.vue
+++ b/browser/flagr-ui/src/components/DebugConsole.vue
@@ -128,7 +128,7 @@ export default {
 }
 </script>
 
-<style lang="less">
+<style lang="less" scoped>
 .json-editor {
   margin-top: 3px;
   .jsoneditor {
--- a/browser/flagr-ui/src/components/Flag.vue
+++ b/browser/flagr-ui/src/components/Flag.vue
@@ -224,10 +224,15 @@
                     <el-card shadow="hover">
                       <el-form label-position="left" label-width="100px">
                         <div class="flex-row id-row">
-                          <div class="flex-row-left">
-                            <el-tag type="primary" :disable-transitions="true"> Variant ID: <b>{{ variant.id }}</b> </el-tag>
-                          </div>
-                          <div class="flex-row-right">
+                          <el-tag type="primary" :disable-transitions="true"> Variant ID: <b>{{ variant.id }}</b> </el-tag>
+                          <el-input
+                            class="variant-key-input"
+                            size="small"
+                            placeholder="Key"
+                            v-model="variant.key">
+                            <template slot="prepend">Key</template>
+                          </el-input>
+                          <div class="flex-row-right save-remove-variant-row">
                             <el-button slot="append" size="small" @click="putVariant(variant)">
                               Save Variant
                             </el-button>
@@ -236,24 +241,17 @@
                             </el-button>
                           </div>
                         </div>
-                        <el-row :gutter="10">
-                          <el-col :span="12">
-                            <el-input
-                              size="small"
-                              placeholder="Key"
-                              v-model="variant.key">
-                              <template slot="prepend">Key</template>
-                            </el-input>
-                          </el-col>
-                          <el-col :span="12">
+                        <el-collapse class="flex-row">
+                          <el-collapse-item title="Variant attachment" class="variant-attachment-collapsable-title">
+                            <p class="variant-attachment-title">You can add a key-value pair (string -> string)</p>
                             <vue-json-editor
                               v-model="variant.attachment"
                               :showBtns="false"
-                              :ref="'json-editor-' + variant.id"
-                              class="json-editor"
+                              :mode="'code'"
+                              class="variant-attachment-content"
                             ></vue-json-editor>
-                          </el-col>
-                        </el-row>
+                          </el-collapse-item>
+                        </el-collapse>
                       </el-form>
                     </el-card>
                   </div>
@@ -567,7 +565,6 @@ export default {
   },
   data () {
     return {
-      refsIntervalId: false,
       loaded: false,
       dialogDeleteFlagVisible: false,
       dialogEditDistributionOpen: false,
@@ -839,29 +836,15 @@ export default {
     },
     toggleShowMdEditor () {
       this.showMdEditor = !this.showMdEditor
-    },
-    setJsonEdiorModeWhenMounted () {
-      if (!this.$refs) {
-        return
-      }
-
-      const jsonEditorRefs = Object.keys(this.$refs).filter(key => key.startsWith('json-editor'))
-
-      jsonEditorRefs.forEach(ref => {
-        this.$refs[ref][0].editor.setMode('code')
-      })
-
-      clearInterval(this.refsIntervalId)
     }
   },
   mounted () {
     this.fetchFlag()
-    this.refsIntervalId = setInterval(this.setJsonEdiorModeWhenMounted, 50)
   }
 }
 </script>
 
-<style lang="less" scoped>
+<style lang="less">
 h5 {
   padding: 0;
   margin: 10px 0 5px;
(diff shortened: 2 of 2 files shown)


browser/flagr-ui/src/components/DebugConsole.vue
```diff
@@ -128,7 +128,7 @@ export default {
 }
 </script>
 
-<style lang="less">
+<style lang="less" scoped>
```
zhouzhuojie: +1 we can keep this scoped

browser/flagr-ui/src/components/Flag.vue
```diff
@@ -246,12 +246,12 @@
...
-                            <el-input
-                              size="small"
-                              placeholder="{}"
-                              v-model="variant.attachmentStr">
-                              <template slot="prepend">Attachment </template>
-                            </el-input>
+                            <vue-json-editor
+                              v-model="variant.attachment"
+                              :showBtns="false"
+                              :ref="'json-editor-' + variant.id"
```
zhouzhuojie: you can just use `:mode="'code'"`, example from https://github.com/dirkliu/vue-json-editor/blob/master/example/App.vue#L4

browser/flagr-ui/src/components/Flag.vue
```diff
@@ -835,15 +839,29 @@ export default {
     },
     toggleShowMdEditor () {
       this.showMdEditor = !this.showMdEditor
+    },
+    setJsonEdiorModeWhenMounted () {
```
zhouzhuojie: with the mode property, I think you can remove this funtion

browser/flagr-ui/src/components/Flag.vue
```diff
@@ -835,15 +839,29 @@ export default {
...
   },
   mounted () {
     this.fetchFlag()
+    this.refsIntervalId = setInterval(this.setJsonEdiorModeWhenMounted, 50)
   }
 }
 </script>
 
-<style lang="less">
+<style lang="less" scoped>
```
zhouzhuojie: We still need the unscoped style to make sure it can be applied to `.el-*` elements. That's why I originally didn't use scoped here.

browser/flagr-ui/src/components/Flag.vue
```diff
@@ -559,10 +562,12 @@ export default {
     debugConsole: DebugConsole,
     flagHistory: FlagHistory,
     draggable: draggable,
-    MarkdownEditor
+    MarkdownEditor,
+    vueJsonEditor
   },
   data () {
     return {
+      refsIntervalId: false,
```
zhouzhuojie: same here, we can remove it

browser/flagr-ui/src/components/Flag.vue
```diff
@@ -835,15 +839,29 @@ export default {
...
+      jsonEditorRefs.forEach(ref => {
+        this.$refs[ref][0].editor.setMode('code')
+      })
+
+      clearInterval(this.refsIntervalId)
     }
   },
   mounted () {
     this.fetchFlag()
+    this.refsIntervalId = setInterval(this.setJsonEdiorModeWhenMounted, 50)
```
zhouzhuojie: same here, we can remove it

browser/flagr-ui/src/components/Flag.vue
```diff
@@ -246,12 +246,12 @@
                             </el-input>
                           </el-col>
                           <el-col :span="12">
-                            <el-input
-                              size="small"
-                              placeholder="{}"
-                              v-model="variant.attachmentStr">
-                              <template slot="prepend">Attachment </template>
-                            </el-input>
+                            <vue-json-editor
```
zhouzhuojie: I wonder if we can add some title or helper string before the json editor to indicate
- It's for variant attachment
- It only supports string => string

image

I'm not good at UI design, so it's really up to you what to design for the attachment section. Maybe we can hide it with a button if it's empty.

browser/flagr-ui/src/components/DebugConsole.vue
yosyad: Alright.

browser/flagr-ui/src/components/Flag.vue
yosyad: Alright, I will add more informative title.

browser/flagr-ui/src/components/Flag.vue
yosyad: You are right, it much simplier, I will change.

browser/flagr-ui/src/components/Flag.vue
yosyad: I prefer always to scope my components style and to add global styles at root scope. What do you think?

yosyad: Alright

browser/flagr-ui/src/components/Flag.vue
yosyad: What do you think about the following?
image

yosyad: One more adjustment:
image

saary:
@zhouzhuojie @yosyad Together with https://github.com/checkr/flagr/pull/259 we are basically allowing for very good user experience when editing more complex JSON attachment data.

Looking forward to seeing this merged.

browser/flagr-ui/src/components/Flag.vue
zhouzhuojie: The variant id is important, so we can't remove it from the UI. Also the key is duplicated when expanded. Otherwise, lgtm.

yosyad: So I can stay with the second design. Is what do you think?
image

yosyad:
@zhouzhuojie I finished the design. It looks like the following image:
image

Please let me know if there is something more than I can do.

browser/flagr-ui/src/components/Flag.vue
```diff
@@ -236,24 +241,17 @@
...
-                              size="small"
-                              placeholder="{}"
-                              v-model="variant.attachmentStr">
-                              <template slot="prepend">Attachment </template>
-                            </el-input>
-                          </el-col>
-                        </el-row>
+                        <el-collapse class="flex-row">
+                          <el-collapse-item title="Variant attachment" class="variant-attachment-collapsable-title">
+                            <p class="variant-attachment-title">You can add a key-value pair (string -> string)</p>
```
zhouzhuojie: after https://github.com/checkr/flagr/pull/259, we can safely say "You can add JSON in key/value pairs format."

browser/flagr-ui/src/components/Flag.vue
```diff
@@ -236,24 +241,17 @@
...
-                            <el-input
-                              size="small"
-                              placeholder="{}"
-                              v-model="variant.attachmentStr">
-                              <template slot="prepend">Attachment </template>
-                            </el-input>
-                          </el-col>
-                        </el-row>
+                        <el-collapse class="flex-row">
+                          <el-collapse-item title="Variant attachment" class="variant-attachment-collapsable-title">
```
zhouzhuojie: just curious, if variant.attachment is not empty (null or {}), can we automatically show the attachment? Just want to highlight the visibility of non-empty attachment.

browser/flagr-ui/src/components/Flag.vue
yosyad: Fixed.

browser/flagr-ui/src/components/Flag.vue
yosyad: It seems like `element-ui` does not support manually setting the `active` flag, so I think that this is not possible to use out of the box capabilities.

saary:
🎉 Waiting for the official docker hub image :)

PR #259, opened 2019-05-23
Add support for arbitrary JSON in attachments

saary:
Allow storing arbitrary JSON structures within variants' attachment field

## Description
Update the Attachment field type from map[string]string to map[string]interface{}

## Motivation and Context
Enable using flagr as a smart general configuration store and remove hacks that use strings in order to store structured data

## How Has This Been Tested?
The existing tests were updated to accommodate for this change

## Types of changes

## Checklist:

### DIFF
--- a/pkg/handler/crud_test.go
+++ b/pkg/handler/crud_test.go
@@ -768,6 +768,33 @@ func TestCrudVariants(t *testing.T) {
 	})
 	assert.Equal(t, *res.(*variant.PutVariantOK).Payload.Key, "another_control")
 
+	res = c.PutVariant(variant.PutVariantParams{
+		FlagID:    int64(1),
+		VariantID: int64(1),
+		Body: &models.PutVariantRequest{
+			Key: util.StringPtr("another_control"),
+			Attachment: map[string]interface{}{
+				"valid_int_value": 1,
+			},
+		},
+	})
+	assert.Equal(t, *res.(*variant.PutVariantOK).Payload.Key, "another_control")
+
+	res = c.PutVariant(variant.PutVariantParams{
+		FlagID:    int64(1),
+		VariantID: int64(1),
+		Body: &models.PutVariantRequest{
+			Key: util.StringPtr("another_control"),
+			Attachment: map[string]interface{}{
+				"valid_structured_value": map[string]interface{}{
+					"string_value": "string",
+					"int_value":    1,
+				},
+			},
+		},
+	})
+	assert.Equal(t, *res.(*variant.PutVariantOK).Payload.Key, "another_control")
+
 	// step 4. it should be able to delete the variant
 	res = c.DeleteVariant(variant.DeleteVariantParams{
 		FlagID:    int64(1),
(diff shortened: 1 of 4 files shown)


pkg/handler/crud_test.go
```diff
@@ -849,20 +849,6 @@ func TestCrudVariantsWithFailures(t *testing.T) {
 		assert.NotZero(t, *res.(*variant.PutVariantDefault).Payload)
 	})
 
-	t.Run("PutVariant - put invalid attachment", func(t *testing.T) {
```
saary: This test was removed as it is now allowed to store artbirary JSON structures

zhouzhuojie: I think we can just change the test to accept string => int payload.

pkg/mapper/entity_restapi/r2e/r2e.go
```diff
@@ -39,11 +39,7 @@ func MapAttachment(a interface{}) (entity.Attachment, error) {
 			return e, fmt.Errorf("all key/value pairs should be string/string. invalid attachment format %s", spew.Sdump(a))
 		}
 		for k, v := range m {
```
zhouzhuojie: no need to copy the key/value pairs here, you can just do `e=m`, or even just `return m, nil`

pkg/mapper/entity_restapi/r2e/r2e.go
```diff
@@ -39,11 +39,7 @@ func MapAttachment(a interface{}) (entity.Attachment, error) {
 			return e, fmt.Errorf("all key/value pairs should be string/string. invalid attachment format %s", spew.Sdump(a))
```
zhouzhuojie: also, need to change the wording here.

pkg/handler/crud_test.go
saary: Will do, might also add a test for more complex types

pkg/mapper/entity_restapi/r2e/r2e.go
saary: Fixed

pkg/mapper/entity_restapi/r2e/r2e.go
saary: fixed

saary:
@zhouzhuojie Thanks for the feedback.
Fixed all comments and also added a test for a more complex structured attachment.

pkg/mapper/entity_restapi/r2e/r2e.go
```diff
@@ -36,15 +36,9 @@ func MapAttachment(a interface{}) (entity.Attachment, error) {
...
 		if !ok {
-			return e, fmt.Errorf("all key/value pairs should be string/string. invalid attachment format %s", spew.Sdump(a))
-		}
-		for k, v := range m {
-			s, ok := v.(string)
-			if !ok {
-				return e, fmt.Errorf("all key/value pairs should be string/string. invalid attachment format %s", spew.Sdump(a))
-			}
-			e[k] = s
+			return e, fmt.Errorf("Make sure JSON is properly formatted. Invalid attachment format %s", spew.Sdump(a))
```
zhouzhuojie: one last nit, JSON format actually supports things like naked string, number, or null. here we may want to say:

"Make sure JSON is properly formatted into key/value pairs."

saary: No problem, pushing a fix now if you want to take a quick look

Issue #261, opened 2019-05-24
Healthcheck Downloading Gzip file?


Shikkic:
Pinging the healthcheck endpoint does not return the expected JSON response 

## Expected Behavior

Hitting the healthcheck endpoint returns a 200 JSON payload, according to the [docs](https://checkr.github.io/flagr/api_docs/#tag/health).

```json
{
  "message": "string"
}
```

https://try-flagr.herokuapp.com/api/v1/health

## Current Behavior

Hitting https://try-flagr.herokuapp.com/api/v1/health returns a gzipped file. This happens locally, on your hosted heroku, and when I deploy it on AWS.

<img width="554" alt="Screen Shot 2019-05-24 at 2 25 53 PM" src="https://user-images.githubusercontent.com/8765687/58348896-de522980-7e2f-11e9-8926-4722438f0082.png">

PR #262, opened 2019-05-24
Fix health check endpoint

zhouzhuojie:
Fix https://github.com/checkr/flagr/issues/261

The reason is that gzip middleware zips empty response without setting content-type, which was empty set in the health swagger definition. 

This pr adds a JSON representation of the health check status:
```
{
  "status": "OK"
}
```

Which may be extended to other things from flagr server.


docs/home.md
```diff
@@ -37,7 +37,7 @@ Install Dependencies.
 
 - Go
 - Make (for Makefile)
-- Yarn (for building UI)
+- Yarn and NPM (for building UI)
```
zhouzhuojie: Plan to migrate to npm only in the near future.

zhouzhuojie:
Tested locally.

```
curl -i localhost:18000/api/v1/health
HTTP/1.1 200 OK
Content-Type: application/json
Vary: Origin
Date: Fri, 24 May 2019 21:31:09 GMT
Content-Length: 16

{"status":"OK"}

```

PR #268, opened 2019-06-04
Add search bar for filtering flags in home page

yosyad:
## Description
I added a search bar that can help to find a flag throw a big number of flags on the home page.

## Motivation and Context
When the usage of flagr grows, I tend to add more flags for every configuration I have. By doing that, I affect directly the length of the list of flags on the home page. I was trying to suggest a new way to navigate throw all the flags easily.

## How Has This Been Tested?
Tested locally, manually.

## Types of changes

## Checklist:

### DIFF
--- a/browser/flagr-ui/src/components/Flags.vue
+++ b/browser/flagr-ui/src/components/Flags.vue
@@ -136,8 +136,8 @@ export default {
       if (this.searchTerm) {
         this.filteredFlags = this.flags.filter(
           ({ id, description}) =>
-            id.toString().startsWith(this.searchTerm) ||
-            description.startsWith(this.searchTerm)
+            id.toString().includes(this.searchTerm) ||
+            description.includes(this.searchTerm)
         )
       }
       else {
--- a/browser/flagr-ui/src/components/Flags.vue
+++ b/browser/flagr-ui/src/components/Flags.vue
@@ -29,8 +29,16 @@
             </el-col>
           </el-row>
 
+          <el-row>
+            <el-input
+              placeholder="Search a flag"
+              prefix-icon="el-icon-search"
+              v-model="searchTerm">
+            </el-input>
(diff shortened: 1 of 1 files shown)


zhouzhuojie:
Hey, this is GREAT! 

I've wanted to do something like the search for a while, but didn't decide which way to go, client side or server side search for the UI?

The advantage of server side search is that we get the limit/offset pagination upfront to be scalable into huge amount of flags. And client side search for its simplicity.

yosyad:
I don't know if flagr home page is a good case for pagination and server side filtering, since I cant see a use case of too many flags (more than 1000?).
I guess that there are a lot of ways we can manage huge amount of flags, for example by grouping them or other approach. In my opinion this is a good start that we can continue with, maybe if will be a future requests from other users, we could implement a more sophisticated solution, such as server side filtering or even pagination. Please share your thoughts.

browser/flagr-ui/src/components/Flags.vue
```diff
@@ -118,8 +128,23 @@ export default {
...
+        this.filteredFlags = [...flags];
       }, handleErr.bind(this))
   },
+  watch: {
+    searchTerm: function() {
+      if (this.searchTerm) {
+        this.filteredFlags = this.flags.filter(
+          ({ id, description}) =>
+            id.toString().startsWith(this.searchTerm) ||
+            description.startsWith(this.searchTerm)
```
zhouzhuojie: is it possible to make it a substring search instead of just `startsWith`? At least, the search should be as good as cmd+f from the browser. This is equivalent to "description_like" params, so we can easily migrate the search box to server-side search in the future (if it's necessary).

yosyad: @zhouzhuojie  Alright, done.

PR #271, opened 2019-06-20
Use lower case and computed for search terms

zhouzhuojie:
A quick change to support toLowerCase match and use computed instead of watch.


zhouzhuojie:
cc @yosyad

yosyad:
> cc @yosyad

Seems LGTM

PR #272, opened 2019-06-27
Feature/import and export flag

yosyad:
The changes in this PR are about adding the capability of import/export flags into the system.

## Description
I thought that it can be handy to copy flags from different environments and to enable an easy way to bootstrap the system from flags list.

import flags

## Motivation and Context
I tend to use flagr in two different environments - DEV and PROD. When I need to add a new flag, first I add it to DEV, then create the same flag manually in PROD. I found it useful to automate the process I tend to do and make it easier to use this library.

## How Has This Been Tested?
Locally, manually.

## Types of changes

## Checklist:

### DIFF
--- a/browser/flagr-ui/src/components/FlagHistory.vue
+++ b/browser/flagr-ui/src/components/FlagHistory.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <el-card v-for="diff in diffs" :key="diff.timestamp" class="snapshot-container">
+    <el-card v-for="diff in diffs" :key="diff.newId" class="snapshot-container">
       <div slot="header" class="el-card-header">
         <el-row>
           <el-col :span="14">
--- a/browser/flagr-ui/src/components/Flags.vue
+++ b/browser/flagr-ui/src/components/Flags.vue
@@ -30,7 +30,10 @@
           </el-row>
 
           <el-row>
-            <import-export-flags :flags="this.flags" class="import-export-flags"></import-export-flags>
+            <import-export-flags
+              :flags="this.flags"
+              class="import-export-flags"
+            ></import-export-flags>
           </el-row>
 
           <el-table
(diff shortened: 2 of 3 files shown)


alonp99:
nice!

zhouzhuojie:
Thanks for the PR! I appreciate the effort, but I think assembling too many requests and logic from frontend is no better than exposing the export/import logic via Flagr's API.

My thoughts of Flagr API for export/import
- Frontend code is harder and fragile to test, ideally, two APIs should be enough. `GET /export` and `POST /import`
- Assemble the logic to the backend allows higher abstraction. For example, we can use swagger to clearly define what's the expected format for `POST /import`, and put enough validation code to gate the input.
- It allows not only Flagr's UI to leverage the export/import, but also programmatically for other clients.
- From Flagr's current API, we already have export to sqlite3 and evaliation_cache's JSON format, it's easy to extend to export flags' JSON format. https://checkr.github.io/flagr/api_docs/#tag/export

Also, for critical UI like this, I'd like to have a dedicated page/tab for export and import, a mistake for clicking the import in production can be very dangerous.

@yosyad What do you think?

yosyad:
> Thanks for the PR! I appreciate the effort, but I think assembling too many requests and logic from frontend is no better than exposing the export/import logic via Flagr's API.
> 
> My thoughts of Flagr API for export/import
> 
> * Frontend code is harder and fragile to test, ideally, two APIs should be enough. `GET /export` and `POST /import`
> * Assemble the logic to the backend allows higher abstraction. For example, we can use swagger to clearly define what's the expected format for `POST /import`, and put enough validation code to gate the input.
> * It allows not only Flagr's UI to leverage the export/import, but also programmatically for other clients.
> * From Flagr's current API, we already have exported to sqlite3 and evaliation_cache's JSON format, it's easy to extend to export flags' JSON format. https://checkr.github.io/flagr/api_docs/#tag/export
> 
> Also, for critical UI like this, I'd like to have a dedicated page/tab for export and import, a mistake for clicking the import in production can be very dangerous.
> 
> @yosyad What do you think?

I think that you are right. We can preserve the existing UI, maybe move it to another page, and implement all its logic in backend.

alonp99:
@yosyad Any progress?

Issue #283, opened 2019-08-28
Attempting to `make all` leads to undefined members due to improper capitalization


Cull-Methi:
## Expected Behavior

When performing `make all`, the software should build. 

## Current Behavior

Currently, it's not building for me on a fresh installation of node and go. It looks like it might be some tooling having changed, causing capitalization to have changed? e.g. `FlagIds` vs `FlagIDs`. Looks similar to #106 

```
# github.com/checkr/flagr/pkg/handler
pkg/handler/crud.go:349:39: params.Body.SegmentIds undefined (type *models.PutSegmentReorderRequest has no field or method SegmentIds, but does have SegmentIDs)
pkg/handler/eval.go:50:24: params.Body.FlagIds undefined (type *models.EvaluationBatchRequest has no field or method FlagIds, but does have FlagIDs)
pkg/handler/export.go:18:32: undefined: export.GetExportSqliteParams
pkg/handler/export.go:22:10: undefined: export.NewGetExportSqliteDefault
pkg/handler/export.go:24:9: undefined: export.NewGetExportSqliteOK
pkg/handler/handler.go:94:5: api.ExportGetExportSqliteHandler undefined (type *operations.FlagrAPI has no field or method ExportGetExportSqliteHandler, but does have ExportGetExportSQLiteHandler)
pkg/handler/handler.go:94:37: undefined: export.GetExportSqliteHandlerFunc
Makefile:24: recipe for target 'build' failed
make: *** [build] Error 2
```

I believe this is a result of the line
```
@CGO_ENABLED=1 GO111MODULE=on go build -mod=vendor -o $(PWD)/flagr github.com/checkr/flagr/swagger_gen/cmd/flagr-server 
```

in the Makefile

## Possible Solution

Not sure. Looks similar to #106?

## Steps to Reproduce (for bugs)

1. Fresh `go get -u github.com/checkr/flagr`
2.Have dependencies installed, or install them fresh
3. Attempt to `make all` in `$GOPATH/src/github.com/checkr/flagr`

## Context

Unable to build from source for development purposes

## Your Environment

* Version used (`flagr version`): source, `master`
* Server type and version:
* Operating System and version (`uname -a`): Debian Jessie
* Link to your project:
* Node/npm versions: 10.16.3/6.9.0
* swagger-merger version: 1.4.3

PR #284, opened 2019-08-28
Bump go-swagger

zhouzhuojie:
Bump to latest go-swagger@0.20.1. Fixed some breaking changes. Note that there are no changes to the flagr APIs at all. Just some variables renaming internally.

Fix https://github.com/checkr/flagr/issues/283


Cull-Methi:
Can confirm that this branch fixes #283, I was able to build using this branch. Thanks for the quick response!

Issue #303, opened 2019-11-07
make fails with dependency error


adnaangr:
Running `make` fails with a deps error

```~/code/flagr master
❯ make
Check deps
Installed github.com/go-swagger/go-swagger/cmd/swagger@v0.20.1 to /Users/mohammadadnaanbadr/go/bin/swagger
Installed github.com/codeskyblue/fswatch@v0.0.0-20190409014817-776b613dacea to /Users/mohammadadnaanbadr/go/bin/fswatch
failed to run go get -d github.com/golangci/golangci-lint/cmd/golangci-lint@v1.16.0: exit status 1
go: finding github.com/golangci/golangci-lint/cmd/golangci-lint v1.16.0
go: finding github.com v1.16.0
go: finding github.com/golangci/golangci-lint/cmd v1.16.0
go: finding github.com/golangci v1.16.0
go get: github.com/golangci/golangci-lint@v1.16.0 requires
	github.com/go-critic/go-critic@v0.0.0-20181204210945-ee9bf5809ead: invalid pseudo-version: does not match version-control timestamp (2019-02-10T22:04:43Z)

make: *** [deps] Error 1

~/code/flagr master 6s
❯ go version
go version go1.13 darwin/amd64
```

zhouzhuojie:
Let me know if the fix works for you.

adnaan:
It works. That was superquick. Thank you :)

PR #304, opened 2019-11-07
Fix golangci-lint deps

zhouzhuojie:
- Fix https://github.com/checkr/flagr/issues/303


zhouzhuojie:
In unit tests, the `make deps` passes now.
image

PR #318, opened 2020-01-29
Add better error checking around 404s and actual database errors

### LINKED ISSUE #317: When flagr encounters database errors during the `GetFlag` func it returns 404
## Expected Behavior

When database errors that aren't "row not found" errors, it should return a 5xx error 

## Current Behavior

Currently, in the case that the database is removed from behind a running instance of flagr (network connection, password change, etc) the application returns a 404 to a GET request on `localhost:13480/api/v1/flags/1000`

## Possible Solution

Adjust the code to check for RecordNotFound errors and return 404s in that case; PR incoming

## Steps to Reproduce (for bugs)

1. Instantiate instance of flagr running against database (postgres, in my case)
2. Remove the ability for flagr to connect to the database (stop the instance, change the user password, etc)
3. do a GET on `localhost:13480/api/v1/flags/1000` or any other flag
4. Get a 404, even if the flag does exist (but is just inaccessible)

## Context

Noticed this when loadtesting a cluster in AWS, that when the database would peak its CPU and requests to it timed out that the application would return 404s instead of 5xxs as expeted

## Your Environment

* Version used (`flagr version`): 1.15
* Server type and version: aws/ECS Fargate
* Operating System and version (`uname -a`):
* Link to your project:

Cull-Methi:
## Description

This change does an explicit check for a RecordNotFound error and returns a 404 only in that case; returning 500 in other cases

## Motivation and Context

Fixes #317 

> Currently, in the case that the database is removed from behind a running instance of flagr (network connection, password change, etc) the application returns a 404 to a GET request on localhost:13480/api/v1/flags/1000

## How Has This Been Tested?

Tested locally by starting up an instance of flagr, verifying functionality, removing the backing database and then attempting to get the same flag again.

## Types of changes

## Checklist:

I don't believe there are tests covering this particular kind of failure case in `crud_tests.go` but I might have missed them - please do point it out and I'll add them if so

### DIFF
--- a/pkg/handler/crud.go
+++ b/pkg/handler/crud.go
@@ -152,11 +152,11 @@ func (c *crud) GetFlag(params flag.GetFlagParams) middleware.Responder {
 	// Flag with given ID doesn't exist, so we 404
 	if result.RecordNotFound() {
 		return flag.NewGetFlagDefault(404).WithPayload(
-			ErrorMessage("Unable to find flag %v in the database", params.FlagID))
+			ErrorMessage("unable to find flag %v in the database", params.FlagID))
 	}
 
 	// Something else happened, return a 500
-	if  err := result.Error; err != nil {
+	if err := result.Error; err != nil {
 		return flag.NewGetFlagDefault(500).WithPayload(
 			ErrorMessage("an unknown error occurred while looking up flag %v: %s", params.FlagID, err))
 	}
--- a/pkg/handler/crud.go
+++ b/pkg/handler/crud.go
@@ -147,10 +147,18 @@ func (c *crud) CreateFlag(params flag.CreateFlagParams) middleware.Responder {
 
 func (c *crud) GetFlag(params flag.GetFlagParams) middleware.Responder {
 	f := &entity.Flag{}
-	err := entity.PreloadSegmentsVariants(getDB()).First(f, params.FlagID).Error
-	if err != nil {
+	result := entity.PreloadSegmentsVariants(getDB()).First(f, params.FlagID)
+
+	// Flag with given ID doesn't exist, so we 404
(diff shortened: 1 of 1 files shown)


pkg/handler/crud.go
```diff
@@ -147,10 +147,18 @@ func (c *crud) CreateFlag(params flag.CreateFlagParams) middleware.Responder {
...
+	if result.RecordNotFound() {
 		return flag.NewGetFlagDefault(404).WithPayload(
-			ErrorMessage("cannot find flag %v. %s", params.FlagID, err))
+			ErrorMessage("Unable to find flag %v in the database", params.FlagID))
+	}
+
+	// Something else happened, return a 500
+	if  err := result.Error; err != nil {
+		return flag.NewGetFlagDefault(500).WithPayload(
+			ErrorMessage("An unknown error occurred while looking up flag %v: %s", params.FlagID, err))
```
zhouzhuojie:
```suggestion
			ErrorMessage("an unknown error occurred while looking up flag %v: %s", params.FlagID, err))
```

go prefers starting with a lower case error message

Cull-Methi:
Do you have any advice on writing a test case for this? Is it possible to remove access to the database during the test runs?

zhouzhuojie:
> Do you have any advice on writing a test case for this? Is it possible to remove access to the database during the test runs?

I think similar to https://github.com/checkr/flagr/blob/master/pkg/handler/crud_test.go#L145, you can have a subtest that stub again for `getDB()`. Not looking close enough though, stub and mock gorm may be tricky.

zhouzhuojie:
LGTM, I will merge this and I can add a simple test in the following PR https://github.com/checkr/flagr/pull/337

PR #343, opened 2020-04-03
Fix enabled query param

### LINKED ISSUE #342: Find Flags enabled=false returning Enabled Flags
For some reason, when calling `/api/v1/flags/?enabled=false` Flags that are enabled are being returned.

Setting `enabled=true` works as expected, but false seems to be off.

## Expected Behavior
Calling the Find Flags API with `enabled=false` in the query string should only return flags that are disabled.

## Current Behavior
Calling the Find Flags API with `enabled=false` in the query string returns all flags.

## Your Environment
Flagr v1.1.7 hosted in Red Hat OpenShift (though I also found the same issue in local docker).

zhouzhuojie:
## Description
Fix #342 

## How Has This Been Tested?
Tested locally.

- `/api/v1/flags` returns all
- `/api/v1/flags?enabled=true` returns enabled ones
- `/api/v1/flags?enabled=false` returns disabled ones

I think the boolean parameter binding from go-swagger also allows `?enabled=t` or `?enabled=f`.  

## Types of changes

## Checklist:

### DIFF
--- a/pkg/handler/crud.go
+++ b/pkg/handler/crud.go
@@ -75,7 +75,7 @@ func (c *crud) FindFlags(params flag.FindFlagsParams) middleware.Responder {
 	q := entity.Flag{}
 
 	if params.Enabled != nil {
-		q.Enabled = *params.Enabled
+		tx = tx.Where("enabled = ?", *params.Enabled)
 	}
 	if params.Description != nil {
 		q.Description = *params.Description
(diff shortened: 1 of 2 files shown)


pkg/handler/crud.go
```diff
@@ -75,7 +75,7 @@ func (c *crud) FindFlags(params flag.FindFlagsParams) middleware.Responder {
 	q := entity.Flag{}
 
 	if params.Enabled != nil {
-		q.Enabled = *params.Enabled
+		tx = tx.Where("enabled = ?", *params.Enabled)
```
zhouzhuojie, line 78: explanation from gorm https://gorm.io/docs/query.html#Struct-amp-Map

PR #348, opened 2020-04-07
Support additional characters for flag keys

### LINKED ISSUE #347: [feat] Support additional characters in keys
In our particular scenario we use a `:` to denote a separation of concern in features. In other cases we also use a `.` notation.

e.g. feature=`organization:create` or feature=`auth.register`

It'd be nice if Flagr could support these same kind of naming conventions, as I'm guessing there's no actual reason not prevent those two characters in key names currently.

dcramer:
Add support for `:` and `.` characters.

Fixes #347

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -12,7 +12,7 @@ import (
 
 var (
 	keyLengthLimit = 63
-	keyRegex       = regexp.MustCompile(`^[\w\d-/]+$`)
+	keyRegex       = regexp.MustCompile(`^[\w\d-/\.\:]+$`)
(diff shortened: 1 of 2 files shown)


dcramer:
Unrelated - how do I avoid go.sum being updated when running make deps/test?

(I havent written go in quite some time so not familiar w/ current ecosystem/tooling)

zhouzhuojie:
LGTM.

For go.sum, if you only run `make deps` or `make vendor` it shouldn't change go.sum, because it serves as a lock with vendors' SHA.

PR #349, opened 2020-04-07
Improve UI consistency and contrast

### LINKED ISSUE #344: Higher contrast color scheme
It'd be great if the color scheme could have better contrast. A light blue background with white text is quite difficult to read.

Happy to @ some designers if thats valuable :)

dcramer:
- Use dark text on light blue background
- Lessen font size of card headers
- Align navbar header elements to page margins
- Improve spacing on Flag Notes
- Format with prettier
- Add `watch` helper

Fixes #344

### DIFF
--- a/docs/home.md
+++ b/docs/home.md
@@ -45,14 +45,20 @@ Build from source.
 # get the source
 go get -u github.com/checkr/flagr
 
-# install dependencies, generated code, and run the app
+# install dependencies, generated code, and start the service in
+# development mode
 cd $GOPATH/src/github.com/checkr/flagr
-make all
+make build start
 ```
 
-If you're modifying the UI, you want may to enable automatic build-on-change:
+If you just want to run the pre-built backend (without the UI development service):
 
 ```
-cd browser/flagr-ui/
-npm run watch
+make run
+```
+
+And alternatively to just run the UI service:
+
+```
+make run_ui
 ```
--- a/docs/home.md
+++ b/docs/home.md
@@ -6,7 +6,6 @@ Flagr is an open source Go service that delivers the right experience to the rig
 
 Run directly with docker. https://hub.docker.com/r/checkr/flagr/
 
-
 ```bash
 # Start the docker container
 docker pull checkr/flagr
@@ -49,4 +48,11 @@ go get -u github.com/checkr/flagr
 # install dependencies, generated code, and run the app
 cd $GOPATH/src/github.com/checkr/flagr
 make all
+```
+
+If you're modifying the UI, you want may to enable automatic build-on-change:
+
+```
+cd browser/flagr-ui/
+npm run watch
 ```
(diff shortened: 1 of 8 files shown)


dcramer:
Sorry that this is large - prettier is a dep in the project and it auto formatted the file. I can try to back that out if you want.

image

dcramer:
Also I realize a lot of this is subjective, and its not my project so if you dont like anything let me know!

dcramer:
One concern is if making this change maybe we should try to update the favicon in parallel? And are there other resources that I might have missed? The docs site might be one, though I'd have to dig to see if the main page of thats in this repo (I suspect its part of docs).

dcramer:
Pushed a commit for docs - though not sure how to test (also not sure how the blue background exists on it).

docs/home.md
```diff
@@ -50,3 +49,10 @@ go get -u github.com/checkr/flagr
 cd $GOPATH/src/github.com/checkr/flagr
 make all
 ```
+
+If you're modifying the UI, you want may to enable automatic build-on-change:
+
+```
+cd browser/flagr-ui/
+npm run watch
```
zhouzhuojie, line 57: I think you can also mention that 
- `npm run serve` if people want to run a vue dev server to just serve the frontend. 
- `npm run watch` if people want to see UI changes from flagr 18000 port - just like the production build.

dcramer: is there a recommended way that you've used for dev to work on the UI? i can just document that and simplify it if its complicated?

hoping to keep it as simple and accessible as your `make run` helper

huge +1000 for makefiles as they're great standard helpers to get things goin :)

zhouzhuojie: how about a new make command:

```
watch_ui:                                     
    @echo "Watch ./browser/flagr-ui for UI development ..."
    @cd ./browser/flagr-ui/; npm run watch                                  
```

dcramer: @zhouzhuojie sorry - was more asking around serve. when you use that do you just access the UI on a different port?

zhouzhuojie: I was thinking that `npm run build --watch` was slow, turns out the build only takes about 3s, so I think we can recommend people to use the build with watch, which is closer to the production experience - e.g. you open localhost:18000 and hit browser refresh. 

But there is some magic from vue-cli serve, like hot reload for css and js changes, so if people prefer, they can also run `npm run serve`.

dcramer: I'm going to suggest a small change to this in an hour or so when i get some cycles.

dcramer: I added a top-level npm package which uses a simple library to run multiple processes, it means you can just `npm start` and you'll get the frontend (in serve mode) and backend running.

zhouzhuojie: I learned that `make` supports parallel commands:

```
run:
	@$(PWD)/flagr --port 18000

run_ui:
	@cd ./browser/flagr-ui/; npm run serve

start:
	$(MAKE) -j run run_ui
```

and then just do `make run` starts both.

dcramer: ah good call thats easier

PR #353, opened 2020-04-08
Add basic auth support for interface

jheth:
Add optional basic auth support to keep the interface from being publicly available.

## Description
- Added new environment variables to enable and configure basic authentication. Disabled by default.
- When enabled, adds a new middleware to support basic auth
- Respect whitelist of paths (prefix or exact match)

## Motivation and Context
- This solves the issue of hosting flagr where it is reachable on the public internet but you do not want outside users to access the web interface.

## How Has This Been Tested?
- Deployed to our hosting provider and manually tested basic auth prompt with various env values.

## Types of changes

## Checklist:

### DIFF
--- a/pkg/config/env.go
+++ b/pkg/config/env.go
@@ -193,7 +193,7 @@ var Config = struct {
 	BasicAuthEnabled              bool     `env:"FLAGR_BASIC_AUTH_ENABLED" envDefault:"false"`
 	BasicAuthUsername             string   `env:"FLAGR_BASIC_AUTH_USERNAME" envDefault:""`
 	BasicAuthPassword             string   `env:"FLAGR_BASIC_AUTH_PASSWORD" envDefault:""`
-	BasicAuthPrefixWhitelistPaths []string `env:"FLAGR_BASIC_AUTH_WHITELIST_PATHS" envDefault:"/api/v1/flags,/api/v1/evaluation,/api/v1/evaluation/batch" envSeparator:","`
+	BasicAuthPrefixWhitelistPaths []string `env:"FLAGR_BASIC_AUTH_WHITELIST_PATHS" envDefault:"/api/v1/flags,/api/v1/evaluation" envSeparator:","`
 	BasicAuthExactWhitelistPaths  []string `env:"FLAGR_BASIC_AUTH_EXACT_WHITELIST_PATHS" envDefault:"" envSeparator:","`
 
 	// WebPrefix - base path for web and API
--- a/pkg/config/env.go
+++ b/pkg/config/env.go
@@ -190,9 +190,9 @@ var Config = struct {
 	HeaderAuthUserField string `env:"FLAGR_HEADER_AUTH_USER_FIELD" envDefault:"X-Email"`
 
 	// Authenticate with basic auth
-	BasicAuthEnabled              bool     `env:"BASIC_AUTH_ENABLED" envDefault:"false"`
-	BasicAuthUsername             string   `env:"BASIC_AUTH_USERNAME" envDefault:""`
-	BasicAuthPassword             string   `env:"BASIC_AUTH_PASSWORD" envDefault:""`
+	BasicAuthEnabled              bool     `env:"FLAGR_BASIC_AUTH_ENABLED" envDefault:"false"`
+	BasicAuthUsername             string   `env:"FLAGR_BASIC_AUTH_USERNAME" envDefault:""`
+	BasicAuthPassword             string   `env:"FLAGR_BASIC_AUTH_PASSWORD" envDefault:""`
 	BasicAuthPrefixWhitelistPaths []string `env:"FLAGR_BASIC_AUTH_WHITELIST_PATHS" envDefault:"/api/v1/evaluation,/static" envSeparator:","`
 	BasicAuthExactWhitelistPaths  []string `env:"FLAGR_BASIC_AUTH_EXACT_WHITELIST_PATHS" envDefault:",/" envSeparator:","`
 
--- a/pkg/config/env.go
+++ b/pkg/config/env.go
@@ -193,8 +193,8 @@ var Config = struct {
 	BasicAuthEnabled              bool     `env:"FLAGR_BASIC_AUTH_ENABLED" envDefault:"false"`
 	BasicAuthUsername             string   `env:"FLAGR_BASIC_AUTH_USERNAME" envDefault:""`
 	BasicAuthPassword             string   `env:"FLAGR_BASIC_AUTH_PASSWORD" envDefault:""`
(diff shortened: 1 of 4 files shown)


pkg/config/middleware.go
```diff
@@ -201,14 +206,66 @@ func (a *auth) whitelist(req *http.Request) bool {
...
+
+	for _, p := range a.PrefixWhitelistPaths {
+		if p != "" && strings.HasPrefix(path, p) {
+			return true
+		}
+	}
+	return false
+}
+
+func (a *basicAuth) ServeHTTP(w http.ResponseWriter, req *http.Request, next http.HandlerFunc) {
```
zhouzhuojie, line 253: do you mind adding a unit test for it? thanks

pkg/config/env.go
```diff
@@ -189,6 +189,13 @@ var Config = struct {
 	HeaderAuthEnabled   bool   `env:"FLAGR_HEADER_AUTH_ENABLED" envDefault:"false"`
 	HeaderAuthUserField string `env:"FLAGR_HEADER_AUTH_USER_FIELD" envDefault:"X-Email"`
 
+	// Authenticate with basic auth
+	BasicAuthEnabled              bool     `env:"FLAGR_BASIC_AUTH_ENABLED" envDefault:"false"`
+	BasicAuthUsername             string   `env:"FLAGR_BASIC_AUTH_USERNAME" envDefault:""`
+	BasicAuthPassword             string   `env:"FLAGR_BASIC_AUTH_PASSWORD" envDefault:""`
+	BasicAuthPrefixWhitelistPaths []string `env:"FLAGR_BASIC_AUTH_WHITELIST_PATHS" envDefault:"/api/v1/flags,/api/v1/evaluation,/static" envSeparator:","`
```
zhouzhuojie, line 196: thinking of the default here, do you want to whitelist `/api/v1/flags` by default? Maybe a good choice to keep the same default as JWT

pkg/config/middleware_test.go
```diff
@@ -315,3 +315,75 @@ func TestAuthMiddlewareWithUnauthorized(t *testing.T) {
...
+		defer func() {
+			Config.BasicAuthEnabled = false
+			Config.BasicAuthUsername = ""
+			Config.BasicAuthPassword = ""
+		}()
+
+		hh := SetupGlobalMiddleware(h)
+		res := httptest.NewRecorder()
+		res.Body = new(bytes.Buffer)
+		req, _ := http.NewRequest("GET", "http://localhost:18000/api/v1/flags", nil)
```
zhouzhuojie, line 344: since you whitelist this by default, which means the actual password checking code path is not covered. For this URL, you can have basic auth encoded into the URL.

pkg/config/middleware_test.go
```diff
@@ -315,3 +315,75 @@ func TestAuthMiddlewareWithUnauthorized(t *testing.T) {
...
+				res := httptest.NewRecorder()
+				res.Body = new(bytes.Buffer)
+				req, _ := http.NewRequest("GET", fmt.Sprintf("http://localhost:18000%s", path), nil)
+				hh.ServeHTTP(res, req)
+				assert.Equal(t, http.StatusOK, res.Code)
+			})
+		}
+	})
+
+	t.Run("it will return 401 for some paths when enabled", func(t *testing.T) {
```
zhouzhuojie, line 367:
```suggestion
	t.Run("it will return 200 for exact whitelist paths", func(t *testing.T) {
```

pkg/config/middleware_test.go
zhouzhuojie: And also, please add a `401` status code case

jheth: I've pushed an update to include 401

pkg/config/env.go
jheth: I'm trying to figure out the best defaults. My main goal is to protect the web interface and let api calls through. Have a suggestion there?

pkg/config/env.go
```diff
@@ -189,6 +189,13 @@ var Config = struct {
 	HeaderAuthEnabled   bool   `env:"FLAGR_HEADER_AUTH_ENABLED" envDefault:"false"`
 	HeaderAuthUserField string `env:"FLAGR_HEADER_AUTH_USER_FIELD" envDefault:"X-Email"`
 
+	// Authenticate with basic auth
+	BasicAuthEnabled              bool     `env:"FLAGR_BASIC_AUTH_ENABLED" envDefault:"false"`
+	BasicAuthUsername             string   `env:"FLAGR_BASIC_AUTH_USERNAME" envDefault:""`
+	BasicAuthPassword             string   `env:"FLAGR_BASIC_AUTH_PASSWORD" envDefault:""`
+	BasicAuthPrefixWhitelistPaths []string `env:"FLAGR_BASIC_AUTH_WHITELIST_PATHS" envDefault:"/api/v1/flags,/api/v1/evaluation,/api/v1/evaluation/batch" envSeparator:","`
```
zhouzhuojie, line 196: Got it, since basicauth isn't like JWT auth, redirecting to a login site makes no sense

A few comments:
- For the prefix whitelist paths, `/api/v1/evaluation` already covers `/api/v1/evaluation/batch`, so you can keep the first.
- Can you briefly document the basic auth behavior for the default setting of BasicAuthPrefixWhitelistPaths and BasicAuthExactWhitelistPaths? For example, the UI access will promote a username/password login, and api will be fine to access the flags (note that it doesn't prevent people from directly curling /api/v1/flags to update flags)

marceloboeira:
PS: You could also do that on your http LB, either nginx or aws ELB, or k8s... it's much easier...

jheth:
Agreed, that would be the easier path. We have this on heroku though which always has a public reachable domain. *.herokuapp.com

marceloboeira:
@jheth I see, yeap, Heroku is extremely customizable on the infra-level. I think you could run `kong` on Heroku and do that, but then it would be more complex eheh

pkg/config/middleware_test.go
```diff
@@ -315,3 +315,62 @@ func TestAuthMiddlewareWithUnauthorized(t *testing.T) {
...
+			Config.BasicAuthUsername = ""
+			Config.BasicAuthPassword = ""
+		}()
+
+		hh := SetupGlobalMiddleware(h)
+		res := httptest.NewRecorder()
+		res.Body = new(bytes.Buffer)
+		req, _ := http.NewRequest("GET", "http://localhost:18000/api/v1/flags", nil)
+		hh.ServeHTTP(res, req)
+		assert.Equal(t, http.StatusOK, res.Code)
```
zhouzhuojie, line 351: actually one more thing to test - a happy code path that the basicauth is serving the traffic with username/password passed. this test didn't cover that because `/api/v1/flags` was one of the default whitelist.

That's why this line wasn't covered yet. https://codecov.io/gh/checkr/flagr/pull/353/diff?src=pr&el=tree#D2-266

jheth: Good call. Thanks for catching that. It's been added.

PR #354, opened 2020-04-09
Add flow to create simple boolean flag

### LINKED ISSUE #345: [feat] Support variant-less rollouts
In contrasting open source feature systems flagr is certainly one of the more mature options. The one caveat I found was its quite complex for a generic boolean feature flag. The scenarios that I think could be simpler are:

- global on/off: could respect the default on/off state.
- % of entity_id on/off: could respect the segment rollout with no distribution plan

You could achieve the latter if the API returned some kind of information saying "segment is on with rollout plan and no variant".

dcramer:
This exposes a way to create a new pre-filled simple flag. It includes a single segment, variant, and predefined distribution allowing you to more easily create a yes/no feature flag check.

It's exposed using the ``{template: "simple"}`` parameter on the flag creation endpoint. Additionally it exposes the action in the UI as a dropdown alongside the existing Create Flag action.

image

Fixes #345


dcramer:
Feel free to nitpick the tests/go/etc here as I just threw stuff together. Haven't ever touched swagger or written Go in years.

Also if there are more descriptive/obvious names we can use for 'template' or 'simple' its worth thinking about that.

go.sum
```diff
@@ -21,18 +21,21 @@ github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf/go.mod h1:ybxpYRF
 github.com/apache/thrift v0.12.0/go.mod h1:cp2SuWMxlEZw2r+iP2GNCdIi4C1qmUzdZFSVb+bacwQ=
 github.com/asaskevich/govalidator v0.0.0-20180315120708-ccb8e960c48f h1:y2hSFdXeA1y5z5f0vfNO0Dg5qVY036qzlz3Pds0B92o=
 github.com/asaskevich/govalidator v0.0.0-20180315120708-ccb8e960c48f/go.mod h1:lB+ZfQJz7igIIfQNfa7Ml4HSf2uFQQRzpGGRXenZAgY=
+github.com/auth0/go-jwt-middleware v0.0.0-20170425171159-5493cabe49f7 h1:irR1cO6eek3n5uquIVaRAsQmZnlsfPuHNz31cXo4eyk=
```
dcramer, line 24: im guessing these are all ok, but every time I run builds it re-adds them to go.sum, so maybe worth just sorting out/committing

browser/flagr-ui/src/components/Flags.vue
```diff
@@ -11,19 +11,26 @@
         <div v-if="loaded">
           <el-row>
             <el-col>
-              <el-input
-                placeholder="Specific new flag description"
-                v-model="newFlag.description">
+              <el-input placeholder="Specific new flag description" v-model="newFlag.description">
```
dcramer, line 14: lots of changes in here from prettier, but most of this is just the onCommandDropdown handler and related elements

pkg/handler/crud.go
```diff
@@ -127,12 +127,56 @@ func (c *crud) CreateFlag(params flag.CreateFlagParams) middleware.Responder {
...
+
+	tx := getDB().Begin()
+
+	if err := tx.Create(f).Error; err != nil {
+		tx.Rollback()
 		return flag.NewCreateFlagDefault(500).WithPayload(
 			ErrorMessage("cannot create flag. %s", err))
 	}
 
+	if params.Body.Template == "simple" {
```
dcramer, line 139: I was hoping there might be a better way to use a constant here, but got a bit overwhelmed when going down that route and figuring out how it'd make sense w/ swagger/etc.

pkg/handler/crud.go
```diff
@@ -142,6 +186,7 @@ func (c *crud) CreateFlag(params flag.CreateFlagParams) middleware.Responder {
 	resp.SetPayload(payload)
 
 	entity.SaveFlagSnapshot(getDB(), f.ID, getSubjectFromRequest(params.HTTPRequest))
```
dcramer, line 188: should this be part of the transaction somehow? just mimicking other endpoints right now but maybe something for followup

pkg/handler/crud_test.go
```diff
@@ -42,6 +42,36 @@ func TestCrudFlags(t *testing.T) {
...
+	t.Run("it should be able to create one simple flag", func(t *testing.T) {
+		res = c.CreateFlag(flag.CreateFlagParams{
+			Body: &models.CreateFlagRequest{
+				Description: util.StringPtr("simple flag"),
+				Key:         "simple_flag_key",
+				Template:    "simple",
+			},
+		})
+		assert.NotZero(t, res.(*flag.CreateFlagOK).Payload.ID)
+		assert.Equal(t, "simple_flag_key", res.(*flag.CreateFlagOK).Payload.Key)
```
dcramer, line 54: I could assert more on the return payload here, as I realized it should return variants/segments as part of it meaning we can do full validation

pkg/handler/crud_test.go
```diff
@@ -42,6 +42,36 @@ func TestCrudFlags(t *testing.T) {
...
+		flagID := uint(res.(*flag.CreateFlagOK).Payload.ID)
+		segment := entity.Segment{FlagID: flagID}
+		db.First(&segment)
+		assert.NotZero(t, segment.ID)
+		assert.Equal(t, segment.Rank, entity.SegmentDefaultRank)
+
+		variant := entity.Variant{FlagID: flagID}
+		db.First(&variant)
```
dcramer, line 62: would have preferred to do .Find() and then assert, but couldnt actually figure out how to construct the query/objects. also didnt read any kind of docs so maybe its not that hard ;)

dcramer:
question for anyone paying attention: are there other kinds of templates that could exist? is that kind of behavior something that might be valuable in heavy use of this system?

pkg/handler/crud_test.go
dcramer: actually looks like payload isnt correct - it does save segment/variant/etc but doesnt return in payload.. will explore but if you have guidance lmk

dcramer: fixed this but would appreciate feedback if the way im managing the arrays is correct

pkg/handler/crud.go
```diff
@@ -127,12 +127,60 @@ func (c *crud) CreateFlag(params flag.CreateFlagParams) middleware.Responder {
...
+	if params.Body.Template == "simple" {
+		// Create our default segment
+		s := &entity.Segment{}
+		s.FlagID = f.ID
+		s.RolloutPercent = uint(100)
+		s.Rank = entity.SegmentDefaultRank
+
+		if err := tx.Create(s).Error; err != nil {
+			tx.Rollback()
+			return segment.NewCreateSegmentDefault(500).WithPayload(ErrorMessage("%s", err))
```
dcramer, line 148: reading this i think these actually need to be all NewCreateFlagDefault

go.sum
zhouzhuojie: yes, usually they are ok, especially when it passes the unit/integration tests.

docs/api_docs/bundle.yaml
```diff
@@ -863,6 +863,9 @@ definitions:
       key:
         description: unique key representation of the flag
         type: string
+      template:
```
zhouzhuojie, line 866: this bundle.yaml is generated, I think can change this file and then `make gen` https://github.com/checkr/flagr/blob/848bcf07b8ff4f6a4a056794d521c62535b2e3b8/swagger/index.yaml#L140-L145

pkg/handler/crud.go
zhouzhuojie: Agreed, I think swagger is more for schema, inside backend, we can use constants as validation. Actually, it's also a good opportunity to do some refactoring, this function is too long. We can have a file `crud_flag_creation.go` in the same handler folder, and dedicates the creation by template logic there. Also the template name `simple`, variant `on` key can be defined as constants.

For the naming of template, how about we start with something explicit first `simple_boolean_flag` instead of `simple`? It's easier to extend that to `simple_ab_experiment` and `simple_dynamic_configuration` in the future.

pkg/handler/crud.go
zhouzhuojie: yes, tracked here https://github.com/checkr/flagr/issues/355

pkg/handler/crud_test.go
zhouzhuojie: Yes, `First` works, since you are finding the first variant with the flagID you just created. Some syntax for query. https://gorm.io/docs/query.html

dcramer:
Thanks! Updated based on feedback.

PR #356, opened 2020-04-14
[WIP] AuthZ in flagr

wildefires:
## Description

This PR is, when complete, meant to introduce AuthZ to flagr. It does not implement AuthN

## Motivation and Context

Allows users to lock down flags from unauthorized users making changes. While changes within an organization are almost certainly well-intentioned, not everyone understands all the wide-reaching effects a change to a configuration can have. This allows flag owners to have more control over who can change it.

I've introduced the concept of authorized users and groups. Non-authorized users will be able to view, but not change fields

## Types of changes

## Checklist:


Cull-Methi:
This is based on a relatively old version of master (sometime early last week, before @dcramer 's additions) thus the merge conflict.

@zhouzhuojie curious of your thoughts on approach for this

Cull-Methi:
image

This is the card that is added just under the base information

And an example of a snapshot where one of the fields changes... 

image

Cull-Methi:
On our internal fork we plan to use an internal auth mechanism that we have for this. What kind of "backend" makes sense for the "default" out of the box solution?

I was thinking defining a yaml file of groups and having the username / groups passed in via the Header. Not sure how to make that happen with the UI, however

zhouzhuojie:
Thanks, what do you think of https://casbin.org/? I'm thinking if we can generalize the authz so that it works for many use cases, and the authz policy itself can be decoupled as a config setting.

pacoguzman:
Would be great if we can provide some kind of interface for this so we can swap to use other implementations.

PR #364, opened 2020-05-07
Oidc key discovery support

dseabolt:
Adding support for validating JWTs from OAuth2 providers using OIDC key discovery. 

## Description
A very popular way of getting JWTs to authenticate users are to get them from OAuth2 providers, and many use OIDC these days. OIDC is a protocol to discover the public signing keys for a JWT so that they can rotate/remove signing keys should they become compromised. This update extends the validation key getter methods to do this key discovery, per the instructions from the root package. 

## Motivation and Context
Allows JWTs issued by OAuth2 providers that use OIDC to be used and validated by Flagr.

## How Has This Been Tested?
I have tested this using Azure Active Directory issued JWTs, and with Okta, a very popular OAuth2 provider used by companies to internally lock down access. 

## Types of changes

## Checklist:
- [ ] My code follows the code style of this project. (I'm unsure about this, I'm relatively new to Go)
- [ ] I have added tests to cover my changes. (I'm not sure how to unit test this since any statically created JWT would have it's signing key potentially removed over time)

### DIFF
--- a/pkg/config/middleware.go
+++ b/pkg/config/middleware.go
@@ -160,33 +160,28 @@ func setupJWTAuthMiddleware() *jwtAuth {
 			}
 
 			// First we need to access the well known url to find out what the jwk_uri is.
-			logrus.Printf("Fetching Oidc Configuration...")
 			resp, err := http.Get(Config.JWTAuthOIDCWellKnownURL)
 			if err != nil {
 				return "", err
 			}
 
 			// Read in the oidc config information, and unmarshal it into our oidc config.
-			logrus.Printf("Reading Oidc Configuration...")
 			defer resp.Body.Close()
 			body, err := ioutil.ReadAll(resp.Body)
 			if err != nil {
 				return "", err
 			}
 
-			logrus.Printf("Parsing Oidc Configuration...")
 			type OidcConfiguation struct {
 				JwkURI string `json:"jwks_uri"` // This is the only field we care about for now.
 			}
 			var oidcConfig OidcConfiguation
 			json.Unmarshal([]byte(body), &oidcConfig)
-			logrus.Printf("OIDC Config: " + string(body))
 
 			if oidcConfig.JwkURI == "" {
 				return "", errors.New("OIDC configuration didn't contain a valid jwks_uri")
 			}
 
-			logrus.Printf("Fetching Oidc Jwks...")
 			// Now we need to read in the jwks info.
 			oidcJwksResp, err := http.Get(oidcConfig.JwkURI)
 			if err != nil {
@@ -197,7 +192,6 @@ func setupJWTAuthMiddleware() *jwtAuth {
 				return "", err
 			}
 
-			logrus.Printf("Reading Oidc Jwks...")
 			// Read in the jwks and unmarshall it.
 			defer oidcJwksResp.Body.Close()
 			body, err = ioutil.ReadAll(oidcJwksResp.Body)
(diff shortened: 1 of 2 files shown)


pkg/config/middleware.go
```diff
@@ -120,6 +128,162 @@ func setupRecoveryMiddleware() *negroni.Recovery {
 	return r
 }
 
+type OidcConfiguation struct {
+	JwkURI string `json:"jwks_uri"` // This is the only field we care about for now.
```
zhouzhuojie, line 132: any specific reason that the field name is not matching the json field name? i.e.

```
JwksURI string `json:"jwks_uri"`
```

pkg/config/middleware.go
```diff
@@ -120,6 +128,162 @@ func setupRecoveryMiddleware() *negroni.Recovery {
...
+}
+
+type OidcJwk struct {
+	JWKeyID     string   `json:"kid"` // JWTs will have a "kid" that will match one of these
```
zhouzhuojie, line 136: same here for the naming of the fields.

pkg/config/middleware.go
```diff
@@ -120,6 +128,162 @@ func setupRecoveryMiddleware() *negroni.Recovery {
 	return r
 }
 
+type OidcConfiguation struct {
+	JwkURI string `json:"jwks_uri"` // This is the only field we care about for now.
+}
+
+type OidcJwk struct {
```
zhouzhuojie, line 135: Can we add a golang style comment like "OidcJwk is a ..."? Same applies to all public exposed symbols.

pkg/config/middleware.go
```diff
@@ -120,6 +128,162 @@ func setupRecoveryMiddleware() *negroni.Recovery {
...
+	Modulus     string   `json:"n"`   // The modulus
+}
+
+type OidcJwksConfiguration struct {
+	Jwks []OidcJwk `json:"keys"`
+}
+
+func discoverOidcJwk(tokenKeyID string) *OidcJwk {
+	// First we need to access the well known url to find out what the jwk_uri is.
+	resp, err := http.Get(Config.JWTAuthOIDCWellKnownURL)
```
zhouzhuojie, line 148: for reliability, maybe it's better to configure this default http client with at least some Timeout settings. Could be added to future PRs.

pkg/config/middleware.go
```diff
@@ -120,6 +128,162 @@ func setupRecoveryMiddleware() *negroni.Recovery {
...
+	SigningKeys []string `json:"x5c"` // We prefer the x5c since it's easier
+	Exponent    string   `json:"e"`   // The least preferred way is recreating the key through exponent.
+	Modulus     string   `json:"n"`   // The modulus
+}
+
+type OidcJwksConfiguration struct {
+	Jwks []OidcJwk `json:"keys"`
+}
+
+func discoverOidcJwk(tokenKeyID string) *OidcJwk {
```
zhouzhuojie, line 146: idiomatic go may be better to return the error directly here, so that the error can be unit tested and bubble up to the caller so that they can handle it at their disposal. 

```
func discoverOidcJwk(tokenKeyID string) (*OidcJwk, error) {

        ...
	if err != nil {
		return nil, err
	}
}

```

pkg/config/middleware.go
zhouzhuojie: maybe we can have similar structure like https://github.com/square/go-jose/blob/master/jwk.go#L41

zhouzhuojie (commented):
BTW, is it possible to leverage existing oidc golang packages? For example, https://github.com/coreos/go-oidc uses gopkg.in/square/go-jose to handle jwks. We can get latest update and security maintenance from the upstream open-source packages.

pkg/config/middleware.go
dseabolt: Yeah, that makes sense.

dseabolt:
I appreciate the comments, go's not my primary language. I considered using the package that you provided, but ultimately all I wanted was the verify part since Flagr's only validating a token's signing, not performing its own sign in flows.

PR #365, opened 2020-05-12
Add sentry env configuration

zhouzhuojie:
## Description
Introduce the configuration for sentry environment setting. In the future, may want to migrate from `raven-go` to `sentry-go`.

- Added `FLAGR_SENTRY_ENVIRONMENT`
- Refactored some config_test.go with defer setup

## How Has This Been Tested?
unit tests and integration tests.

## Types of changes

## Checklist:

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -9,48 +9,57 @@ import (
 
 func TestSetupSentry(t *testing.T) {
 	Config.SentryEnabled = true
-	assert.NotPanics(t, func() {
-		setupSentry()
-	})
-	Config.SentryEnabled = false
+	Config.SentryEnvironment = "test"
+	defer func() {
+		Config.SentryEnabled = false
+		Config.SentryEnvironment = ""
+	}()
+
+	assert.NotPanics(t, func() { setupSentry() })
 }
 
 func TestSetupNewRelic(t *testing.T) {
 	Config.NewRelicEnabled = true
-	assert.Panics(t, func() {
-		setupNewrelic()
-	})
-	Config.NewRelicEnabled = false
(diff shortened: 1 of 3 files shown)


pkg/config/config_test.go
```diff
@@ -9,48 +9,57 @@ import (
```
zhouzhuojie, line 9: Just did some refactor and use `defer` pattern for config reset in the tests.

pkg/config/env.go
```diff
@@ -72,8 +72,9 @@ var Config = struct {
 	CORSEnabled bool `env:"FLAGR_CORS_ENABLED" envDefault:"true"`
 
 	// SentryEnabled - enable Sentry and Sentry DSN
-	SentryEnabled bool   `env:"FLAGR_SENTRY_ENABLED" envDefault:"false"`
-	SentryDSN     string `env:"FLAGR_SENTRY_DSN" envDefault:""`
+	SentryEnabled     bool   `env:"FLAGR_SENTRY_ENABLED" envDefault:"false"`
+	SentryDSN         string `env:"FLAGR_SENTRY_DSN" envDefault:""`
```
ivanRylach, line 76: Why do we prefix these env vars with `FLAGR_`? They seem pretty standard.

zhouzhuojie: It's optional, and in case flagr was deploying together with other apps in the same container, it's also good to isolate/prefix the environment variables just for the open source app.

People can definitely set the official `SENTRY_ENVIRONMENT` and leave `FLAGR_SENTRY_ENVIRONMENT` empty if they already have that configured. That will be respected by the official raven/sentry package.

Issue #362, opened 2020-04-28
Add Tags for Organizational Management


bradphilips:
At a large organization like Progressive, we need a way of easily identifying Flags that are aligned with certain organizations (e.g. Mobile, Online Servicing) as well as separate types of tags such as Feature Flags, A/B Tests and Configuration Distribution. Tags is a feature that is commonly used to organize and provide searchability across large groupings of like objects, similar to #hashtags on social media or file tags in macOS Finder. This works by assigning descriptive tag(s) to these objects so looking for them is as simple as looking for the tag and anything associated to it will be found. We would like to add a tagging system to Flagr to accomplish the same goals.
 
This could be used to identify Flags easily by searching their assigned tags and batch evaluate them all in one call.
 
Outside of this, we could also use this for organization of flags of different types or assigned to different organizations, and potentially for reporting and access control.
 
**Design:**
 
How we are thinking of this from a design perspective is that Tags would be a many to many relationship to Flags. The tasks we have identified are as follows:
 
1. Create a Tag entity with the following schema:
 
```go
type Tag struct {
	gorm.Model

	Value string  `sql:"type:text" gorm:"unique;not null"`
	Flags []*Flag `gorm:"many2many:flags_tags"`
}
```

2. Modify the Flag.go to reference a many to many relationship between Tags:
 
```go
// Flag is the unit of flags
type Flag struct {
	gorm.Model

	Key         string `gorm:"type:varchar(64);unique_index:idx_flag_key"`
	Description string `sql:"type:text"`
	CreatedBy   string
	UpdatedBy   string
	Enabled     bool
	Segments    []Segment
	Variants    []Variant
	**Tags        []Tag `gorm:"many2many:flags_tags;"`**
	SnapshotID  uint
	Notes       string `sql:"type:text"`

	DataRecordsEnabled bool
	EntityType         string

	FlagEvaluation FlagEvaluation `gorm:"-" json:"-"`
}
```
 
3. Provide Get, Create and Delete endpoints for tags (/flags/{id}/tags, /flags/{id}/tags/{id}
4. Modify the FindFlags to allow searching for Tags
5. Modify the Evaluate Batch to evaluate by a flagTags property
6. Modify the UI to provide the ability to add tags to flags
7.  Modify the search to be able to search Flags by Tag
8. Add Unit Tests
9. Add endpoints for managing Tags independently? (e.g. /tags, /tags/{id})
  a. We’re not sure this is needed – let us know your thoughts
 
[...]

PR #369, opened 2020-06-02
Added Tags for Organizational Management

fenriskiba:
# Added Tags for Organizational Management

Developed in collaboration with @bradphilips and @nambup.

## Description
* Adding Tags in a many-to-many relationship with flag
* Added tags query parameter to findFlags
* Added findTags, createTag, deleteTag, findAllTags API calls to manage Tags
* Added option to postEvaluationBatch to evaluate for all Flags with provided Tags
* Added unit and integration tests for Tags
* Added elements in the UI to search, edit, and view Tags on a Flag

## Motivation and Context
Please see Issue #362

## How Has This Been Tested?
* Unit Tests
* Integration Tests
* Tested in Local
* Tested in internal sandbox environment

## Types of changes

## Checklist:


browser/flagr-ui/src/components/Flags.vue
```diff
@@ -123,15 +124,21 @@ export default {
...
+              tags.map(tag => tag.value.toLowerCase().includes(term.toLowerCase())).includes(true)
+            }).every((e) => e)
         );
       }
       return this.flags;
     }
   },
   methods: {
+    tagFormatter(row, col, val) {
+      return val.map(tag => tag.value).join(', ');
```
zhouzhuojie, line 140: maybe we can use el-tag for showing tags

image

browser/flagr-ui/src/components/Flag.vue
```diff
@@ -257,6 +257,36 @@
...
+                        @close="deleteTag(tag)">
+                        {{tag.value}}
+                      </el-tag>
+                      <el-autocomplete
+                        class="tag-key-input"
+                        v-if="tagInputVisible"
+                        v-model="newTag.value"
+                        ref="saveTagInput"
+                        size="mini"
+                        :trigger-on-focus="false"
```
zhouzhuojie, line 280: I think if we set trigger-on-focus to true, it helps to standardize tags creation and reduce typos

browser/flagr-ui/src/components/Flag.vue
```diff
@@ -896,6 +934,60 @@ export default {
...
+      this.newTag = clone(DEFAULT_TAG);
+      this.tagInputVisible = false;
+    },
+    queryTags(queryString, cb) {
+      let results = this.allTags.filter((tag) => tag.value.toLowerCase().includes(queryString.toLowerCase()));
+      cb(results);
+    },
+    loadAllTags() {
+      Axios.get(
+        `${API_URL}/tags`
```
zhouzhuojie, line 962: I noticed that even when we delete the tag from across flags, it still shows up in the autocomplete.

pkg/handler/eval_cache.go
```diff
@@ -59,6 +63,39 @@ func (ec *EvalCache) Start() {
 	}()
 }
 
+func (ec *EvalCache) GetByTags(tags []string) []*entity.Flag {
+	ec.mapCacheLock.RLock()
+	defer ec.mapCacheLock.RUnlock()
+
+	results := []*entity.Flag{}
+	for _, t := range tags {
+		s := util.SafeString(t)
```
zhouzhuojie, line 72: t is already a string type, so no need to cast to string here

pkg/handler/eval_cache.go
```diff
@@ -59,6 +63,39 @@ func (ec *EvalCache) Start() {
...
+		f, ok := ec.tagCache[s]
+		if ok {
+			results = append(results, f...)
+		}
+	}
+
+	return FlattenFlags(results)
```
zhouzhuojie, line 79: I can see that you are doing dedup here. Since flag id order doesn't matter, how about we define 

```
tagCache:        map[string]map[uint]*entity.Flag      // i.e. tag => map[flagID]*entity.Flag
```

And when loop through the tags, we just need to get a union of the maps, and then return the values of that union map as a slice

pkg/handler/eval.go
```diff
@@ -39,7 +39,8 @@ func (e *eval) PostEvaluation(params evaluation.PostEvaluationParams) middleware
 			ErrorMessage("empty body"))
 	}
 
-	evalResult := EvalFlag(*evalContext)
+	flag := LookupFlag(*evalContext)
+	evalResult := EvalFlagWithContext(flag, *evalContext)
```
zhouzhuojie, line 43: since the function EvalFlag is still there and `flag` variable is not used below the line, I think you can keep `evalResult := EvalFlag(*evalContext)`

pkg/entity/tag.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"github.com/jinzhu/gorm"
+)
+
+// Tag is a descriptive identifier given to ease searchability
+type Tag struct {
+	gorm.Model
+
+	Value string  `sql:"type:varchar(64);unique_index:idx_tag_value"`
```
zhouzhuojie, line 11: Thinking about the validation of `Value` field, I think it's probably safer to start with stricter regex pattern, as the tag may grow with inconsistent naming pattern. e.g. `eng_team_abc` vs `Eng Team EFG`. A good starting point is `util.IsSafeKey`.

In the future, we can decouple the validation of tags with loose restrictions if needed.

browser/flagr-ui/src/components/Flag.vue
bradphilips: @zhouzhuojie this was set to false because focus was triggering a save operation which we didn't was causing the saveTagInput method such that it ignores values that are empty but would still cancel your operation instead. This would also conflict with entry process we're using with the save on enter and cancel.

browser/flagr-ui/src/components/Flags.vue
bradphilips: Yeah we'll work on getting this working 👍

browser/flagr-ui/src/components/Flag.vue
bradphilips: This is actually intended for the Flags delete, because it simply removes the reference to the Tag from the Flag and not the actual Tag itself because it could be referenced by another Flag. Let us know if there's another way you'd like to handle this.

pkg/handler/eval_cache.go
bradphilips: Perfect, we'll remove this 👍

pkg/entity/tag.go
bradphilips: @zhouzhuojie Makes perfect sense, could `util.IsSafeKey` be re-used here or should we develop an alternate method with our own requirements?

bradphilips:
@zhouzhuojie Made some changes for items 1,4,5 and 6. Let us know what you'd like to do on the others. Thanks!

pkg/entity/tag.go
bradphilips: Added validation for this in the last commit. I modified the regex from the `IsSafeKey` slightly to allow for spaces. Let me know if this is to your liking. Thanks!

browser/flagr-ui/src/components/Flag.vue
fenriskiba: @zhouzhuojie, I was talking to @bradphilips and a couple of our other Flagr users, and we wondered if this would be a preferred way to handle this anyway. It would keep us from creating and deleting the same tag over and over if it goes in and out of use.

Brad also mentioned an option to search for any flags with the removed tag and fully delete the tag if there aren't any. I can see an argument for doing that to prevent unused tags from sticking around.

Personally, I like leaving them the way it's currently set up, but I don't think we'd argue if you want to fully delete the tags.

docs/api_docs/bundle.yaml
```diff
@@ -709,6 +792,32 @@ paths:
...
           schema:
             $ref: '#/definitions/error'
+  /tags:
+    get:
+      tags:
+        - tag
+      operationId: findAllTags
+      parameters:
+        - in: query
+          name: limit
```
zhouzhuojie, line 802: nit, we may also want to add `offset` to enable pagination

pkg/handler/crud.go
```diff
@@ -259,6 +279,74 @@ func (c *crud) DeleteFlag(params flag.DeleteFlagParams) middleware.Responder {
...
+func (c *crud) DeleteTag(params tag.DeleteTagParams) middleware.Responder {
+	t := &entity.Tag{}
+	t.ID = uint(params.TagID)
+
+	s := &entity.Flag{}
+	s.ID = uint(params.FlagID)
+
+	if err := getDB().Find(s).Association("Tags").Delete(t).Error; err != nil {
+		return tag.NewDeleteTagDefault(500).WithPayload(ErrorMessage("%s", err))
+	}
```
zhouzhuojie, line 291: You may also need `entity.SaveFlagSnapshot()` to capture tag deletion in flag history.

pkg/handler/crud_test.go
```diff
@@ -907,6 +942,105 @@ func TestCrudVariantsWithFailures(t *testing.T) {
...
+	defer db.Close()
+	defer gostub.StubFunc(&getDB, db).Reset()
+
+	c.CreateFlag(flag.CreateFlagParams{
+		Body: &models.CreateFlagRequest{
+			Description: util.StringPtr("funny flag"),
+		},
+	})
+
+	// step 0. it should return 0 tags before creaetion
```
zhouzhuojie, line 959: can you also add some tests for `limit` and `tags like` parameter?

pkg/handler/eval_cache.go
```diff
@@ -59,6 +62,37 @@ func (ec *EvalCache) Start() {
...
+		f, ok := ec.tagCache[t]
+		if ok {
+			results = merge(results, f)
+		}
+	}
+
+	return values(results)
+}
+
+func values(m map[uint]*entity.Flag) []*entity.Flag {
```
zhouzhuojie, line 80: I think the function names `values` and `merge` are too generic inside the handler package, and these two functions are simple enough and I think it's better to just inline into the GetByTags function.

bradphilips:
> Left a few comments, and awesome work! Thanks for the contribution. I will take a look again soon.

Awesome, thank you @zhouzhuojie -- Fixed all the additional comments as well. Let us know if there's anything additional.

bradphilips:
@zhouzhuojie thanks so much for the merge. Do you have an idea of when this will get into a release? We're eager to deploy this in our infrastructure! Thanks!

bradphilips:
> @zhouzhuojie thanks so much for the merge. Do you have an idea of when this will get into a release? We're eager to deploy this in our infrastructure! Thanks!

I see that it was just done. Thank you! :)

Issue #255, opened 2019-05-15
Unable to re-use flag key from deleted flag


kenli-rubrik:
## Expected Behavior
Suppose I specify flag_key="blah" for flag 1. I delete flag 1, and then try to create a new flag with flag_key="blah". I would expect that I can create the new flag successfully since the flag key is no longer being used.

## Current Behavior
Instead, I get the error:
```
Error 1062: Duplicate entry 'thisisatestflag' for key 'idx_flag_key'
```

## Possible Solution
This is likely due to GORM using soft deletes, so we're unable to add the new flag because the key still exists in the index.

Temporary workaround we're planning to use is to modify the flag key to a dummy value before deleting any flags. A real solution could involve detecting that the flag key exists under another flag and either:
1. Stripping the key from the old flag, and continue creating the new flag.
2. Reviving the old flag (and clearing other related data like segments, variants, etc?). Related to https://github.com/checkr/flagr/issues/215

## Steps to Reproduce (for bugs)
1. Create a flag with a know key
2. Delete the flag from step 1
3. Attempt to create a new flag with the same flag key

## Context
We have some testing utils that will create a temporary flag (identified by key), run some code that depends on the flag, and then delete the flag. However consecutive attempts to re-use the same key fails due to this issue.

zhouzhuojie:
I actually like that we keep the unique index including deleted flags, this makes restoring the deleted flags logic simple and clean in the future. 

Is it possible to create temporary flags with a randomly generated key for each test? If not, I would prefer resetting the data storage of flagr so each test you can have a clean setup.

kenli-rubrik:
I think if we had the ability to restore flags, that would be mostly sufficient. As long as we could easily detect that a flag was deleted and revive it instead of creating it again.

To expand on the context a bit more, it's not just testing where this edge case comes up. We also have a script that synchronizes flag state from a YAML file (deleting any flags that no longer exist). If a developer removes a flag and then tries to re-add it later, the script will fail. Or another example is that a flag is added to the YAML file, the commit is reverted, and sometime later the flag is added again.

zhouzhuojie:
Got it, I was thinking of syncing flags recently and added features like exporting the eval cache to JSON files. For example, we are using the exported JSON files in our test and staging environment. It basically makes the flagr instance into eval_only (read-only) mode, so you don't even need to care about deleted flags, since there's no DB and no writes are allowed, and the source of truth will the JSON file. I think it may be one of the options for your setup.

[...]

PR #379, opened 2020-07-22
Allow users the ability to restore deleted flags

### LINKED ISSUE #215: [feat] restore deleted flags
Some flags are sensitive, if it's deleted, make sure it's also easy to restore as well. Currently, in gorm, everything is soft deleted. I think this feature will need the following changes:

- UI front page, to not only show the current flags, but also a section to show deleted flags.
- A new parameter in `GET /flags?deleted=true` to get deleted flags
- A new route `POST /flags/<id>/restore` to restore deleted flags
- A new UI button to restore deleted flags

silent1mezzo:
This allows users to restore delete flags from the main page.

Fixes #215 and partially solves #255 

Screen Shot 2020-07-22 at 2 01 00 PM


silent1mezzo:
This is the first time working in Vue or Go so feel free to point out where I can improve the code.

pkg/handler/crud.go
```diff
@@ -272,6 +279,29 @@ func (c *crud) SetFlagEnabledState(params flag.SetFlagEnabledParams) middleware.
 	return resp
 }
 
+func (c *crud) RestoreFlag(params flag.RestoreFlagParams) middleware.Responder {
+	f := &entity.Flag{}
+	if err := getDB().Unscoped().First(f, params.FlagID).Error; err != nil {
```
zhouzhuojie, line 284: since it's primarily used at the front page with the flags already loaded with tags, I think it's better to preload tags.
```suggestion
	if err := entity.PreloadFlagTags(getDB().Unscoped()).First(f, params.FlagID).Error; err != nil {
```

browser/flagr-ui/src/components/Flags.vue
```diff
@@ -84,6 +84,49 @@
               </template>
             </el-table-column>
           </el-table>
+
+          <h2>Deleted Flags</h2>
```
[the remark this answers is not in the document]
silent1mezzo: @zhouzhuojie sounds good. I just updated it so it hides the Deleted flags and only loads them when a user clicks on the collapsed table.

pkg/handler/crud.go
```diff
@@ -108,6 +110,11 @@ func (c *crud) FindFlags(params flag.FindFlagsParams) middleware.Responder {
 			fmt.Sprintf("%%%s%%", strings.ToLower(*params.DescriptionLike)),
 		)
 	}
+	if params.Deleted != nil {
```
zhouzhuojie, line 113: we also need to check if deleted is true or false.

```suggestion
	if params.Deleted != nil && *params.Deleted {
```

silent1mezzo:
Thanks for the catch on that API endpoint @zhouzhuojie

Issue #386, opened 2020-08-11
MYSQL image compatibility


rsahoo752055:
When we are trying to setup flagr image with mysql database (image version: latest/8/8.0/8.0.21), we are getting below error from mysql image container.
Error:: "INFO[0101] /go/src/github.com/checkr/flagr/pkg/handler/eval_cache_fetcher.go:147Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'ASC,id ASC' at line 1
ERRO[0101] reload evaluation cache error                 err="Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'ASC,id ASC' at line 1".

But when i am trying to setup flagr image with 5.7.31 or below mysql image version, flagr is working as excepted.

Please confirm mysql image compatibility version with flagr.

brasilikum:
Hey,
it seems like it worked as an accident.

Support for mysql is tracked here:
https://github.com/checkr/flagr/pull/331

Duplicate of #326

zhouzhuojie:
> Hey,
> it seems like it worked as an accident.
> 
> Support for mysql is tracked here:
> #331
> 
> Duplicate of #326

Oh, that's for MSSQL. I think this issue is related to MYSQL's compatibility. Will look into it soon.

PR #395, opened 2020-09-05
Fix mysql8 compatibility

zhouzhuojie:
Fix mysql8 compatibility by removing the unnecessary "ASC" keyword in preload gorm queries.
The actual root cause is that `Rank` is a new keyword in mysql8, and we need to safely escape that for MySQL and Postgres.

Tried a few solutions, and the safest is actually remove the "ASC" so that gorm can handle the name escape correctly for mysql and postgres.
 

## Description
Fix https://github.com/checkr/flagr/issues/386

## How Has This Been Tested?
Tested locally for mysql 5.6 and mysql 8+.
Tested in integration tests.

## Types of changes

## Checklist:


integration_tests/docker-compose.yml
```diff
@@ -14,6 +14,14 @@ services:
...
       MYSQL_USER: "root"
 
+  mysql8:
+    image: mysql:8
+    container_name: flagr-mysql8
+    environment:
+      MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
+      MYSQL_DATABASE: "flagr"
+      MYSQL_USER: "root"
+
```
zhouzhuojie, line 24: added integration tests for mysql 5.6 and mysql 8

pkg/entity/flag.go
```diff
@@ -36,23 +36,23 @@ type FlagEvaluation struct {
...
 
 // PreloadSegmentsVariantsTags preloads segments, variants and tags for flag
 func PreloadSegmentsVariantsTags(db *gorm.DB) *gorm.DB {
-	return db.
+	return db.Debug().
 		Preload("Segments", func(db *gorm.DB) *gorm.DB {
 			return PreloadConstraintsDistribution(db).
-				Order("rank ASC").
-				Order("id ASC")
+				Order("rank").
```
zhouzhuojie, line 48: `"rank ASC"` was not escaped correctly in mysql8 due to the new keyword `RANK` function. Here we can just remove `ASC` and let gorm handle the field name escape correctly.

marceloboeira (approved):

PR #408, opened 2020-10-09
Evaluation with FlagTagsOperator to match any or all tags

pacoguzman:
Finding flags to be evaluated using tags allow to pass ANY/ALL tag matching as operator

## Description

Just support a new enum value `ANY` or `ALL` to determine which flags matches the tags the default is `ANY`. `ALL` just filter out results applying tags one after the other.

## Motivation and Context

The main idea of the change is to allow to pass `flagsTagsOperator` on batch evaluation requests so we can handle scenarios where each tag used to describe the flags are independent so we could want to match flags which contains all the tags passed. Currently, tags are considered as any tag in the flags make the match.

We have tags determining the ownership of the flags, and some teams group flags inside other tags and we have some generic rules to naming it.

- ownership -> `owner:team-name`
- batch -> `batch:batch-name`

So as could be collisions on the names between teams we think that adding a match all operator for tags solve this easily to just evaluate on those flags matching all the tags provided. This is opt in as the default behaviour is the same as before.

## How Has This Been Tested?

I've added a couple of tests in the codebase to describe the new behaviour

## Types of changes

## Checklist:

### DIFF
--- a/docs/api_docs/bundle.yaml
+++ b/docs/api_docs/bundle.yaml
@@ -1311,6 +1311,17 @@ definitions:
         x-omitempty: true
         items:
           type: string
+      flagTagsOperator:
+        description: >-
+          determine how flagTags is used to filter flags to be evaluated. OR
+          extends the evaluation to those which contains at least one of the
+          provided flagTags or AND limit the evaluation to those which contains
+          all the flagTags.
+        type: string
+        enum:
+          - ANY
+          - ALL
+        default: ANY
   evalResult:
     type: object
     properties:
@@ -1403,6 +1414,17 @@ definitions:
           type: string
           minLength: 1
         minItems: 1
+      flagTagsOperator:
+        description: >-
+          determine how flagTags is used to filter flags to be evaluated. OR
+          extends the evaluation to those which contains at least one of the
+          provided flagTags or AND limit the evaluation to those which contains
+          all the flagTags.
+        type: string
+        enum:
+          - ANY
+          - ALL
+        default: ANY
   evaluationBatchResponse:
     type: object
     required:
--- a/pkg/handler/eval.go
+++ b/pkg/handler/eval.go
@@ -51,17 +51,19 @@ func (e *eval) PostEvaluationBatch(params evaluation.PostEvaluationBatchParams) 
 	flagIDs := params.Body.FlagIDs
 	flagKeys := params.Body.FlagKeys
 	flagTags := params.Body.FlagTags
+	flagTagsOperator := params.Body.FlagTagsOperator
 	results := &models.EvaluationBatchResponse{}
 
 	// TODO make it concurrent
 	for _, entity := range entities {
 		if len(flagTags) > 0 {
 			evalContext := models.EvalContext{
-				EnableDebug:   params.Body.EnableDebug,
-				EntityContext: entity.EntityContext,
-				EntityID:      entity.EntityID,
-				EntityType:    entity.EntityType,
-				FlagTags:      flagTags,
+				EnableDebug:      params.Body.EnableDebug,
+				EntityContext:    entity.EntityContext,
+				EntityID:         entity.EntityID,
+				EntityType:       entity.EntityType,
+				FlagTags:         flagTags,
+				FlagTagsOperator: flagTagsOperator,
 			}
 			evalResults := EvalFlagsByTags(evalContext)
 			results.EvaluationResults = append(results.EvaluationResults, evalResults...)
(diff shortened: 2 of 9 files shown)


docs/api_docs/bundle.yaml
```diff
@@ -1311,6 +1311,14 @@ definitions:
         x-omitempty: true
         items:
           type: string
+      flagTagsOperator:
```
zhouzhuojie, line 1314: btw, this doc is generated, you can move the change to https://github.com/checkr/flagr/blob/master/swagger/index.yaml#L518, and then run `make gen` to get the same result

pkg/handler/eval.go
```diff
@@ -133,8 +135,14 @@ var LookupFlag = func(evalContext models.EvalContext) *entity.Flag {
...
 	cache := GetEvalCache()
+	var operator string
+	if evalContext.FlagTagsOperator == nil {
+		operator = ""
+	} else {
+		operator = *evalContext.FlagTagsOperator
+	}
 
-	fs := cache.GetByTags(evalContext.FlagTags)
+	fs := cache.GetByTags(evalContext.FlagTags, operator)
```
zhouzhuojie, line 145: for readability, I think you can just pass the `evalContext.FlagTagsOperator` as `*string` to `cache.GetByTags`, and then check if `if operator == nil || *operator == models.EvaluationBatchRequestFlagTagsOperatorANY` there.

pkg/handler/eval_cache.go
```diff
@@ -70,13 +92,43 @@ func (ec *EvalCache) GetByTags(tags []string) []*entity.Flag {
...
+func (ec *EvalCache) getByTagsALL(tags []string) map[uint]*entity.Flag {
+	results := map[uint]*entity.Flag{}
+	cache := ec.cache.Load().(*cacheContainer)
+
+	for i, t := range tags {
+		fSet, ok := cache.tagCache[t]
+		if !ok {
+			// clear as no flags
+			results = map[uint]*entity.Flag{}
+			continue
```
zhouzhuojie, line 107: if the tag is not there, I think should we just return `nil` here given the `ALL` operator.

pkg/handler/eval_cache.go
```diff
@@ -70,13 +92,43 @@ func (ec *EvalCache) GetByTags(tags []string) []*entity.Flag {
...
+		}
+
+		if i == 0 {
+			// store all the flags
+			for fID, f := range fSet {
+				results[fID] = f
+			}
```
zhouzhuojie, line 114: you can just do `results = fSet`

zhouzhuojie: nvm, I think yours is better, since we are deleting the keys from results on the fly, better keep a shallow clone of the map.

pkg/handler/eval_cache.go
```diff
@@ -70,13 +92,43 @@ func (ec *EvalCache) GetByTags(tags []string) []*entity.Flag {
...
+				// accept only previous flags
+				missingFIDs := []uint{}
+				for fID := range results {
+					if fSet[fID] == nil {
+						missingFIDs = append(missingFIDs, fID)
+					}
+				}
+				for _, fID := range missingFIDs {
+					delete(results, fID)
+				}
```
zhouzhuojie, line 126: How about one pass of the keys in results, and just delete the keys not the current fSet?

```
				for fID := range results {
					if _, ok := fSet[fID]; !ok {
						delete(results, fID)
					}
				}
```

docs/api_docs/bundle.yaml
pacoguzman: Sure, I wasn't sure how this work

pkg/handler/eval.go
pacoguzman: Thanks reads better

pkg/handler/eval_cache.go
pacoguzman: Cool, I wasn't sure if in Go is possible to modify a slice while iterating it. Changing it for sure

pkg/handler/eval_cache.go
pacoguzman: You're totally right

pacoguzman: But is it correct to return nil or should we return an empty list?

pacoguzman:
I've applied all your feedback as that makes a lot of sense and simplify the solution. Thanks for the review

pkg/handler/eval_cache.go
```diff
@@ -70,13 +92,37 @@ func (ec *EvalCache) GetByTags(tags []string) []*entity.Flag {
...
+			return map[uint]*entity.Flag{}
+		}
+
+		if i == 0 {
+			// store all the flags
+			for fID, f := range fSet {
+				results[fID] = f
+			}
+		} else {
+			if len(results) > 0 {
```
zhouzhuojie, line 115: one more thing to optimize is that you can return early when `len(results) == 0`

pacoguzman: Done, after each delete loop we check if there is still any match to continue or return early.

pkg/handler/eval_cache.go
zhouzhuojie: related topic https://stackoverflow.com/questions/23229975/is-it-safe-to-remove-selected-keys-from-map-within-a-range-loop, I also learned that `delete` marks the keys as deleted, thus safe to keep iterating.

zhouzhuojie (approved):
🎉

Issue #214, opened 2019-02-06
[feat] add notification channel for flag updates


zhouzhuojie:
Teams need more visibility into how and when flags are updated. For example, Email or Slack notifications can be good channels.

Not sure if this feature is a good addon to the current Flagr setup. Maybe we can build it in Flagr, maybe we can expose Flagr webhooks and let other applications implement it.

crberube:
The webhook approach sounds interesting because that would enable flexibility for many different use cases / notification channels.

PR #410, opened 2020-10-16
(j4k work continuation) Notifier interface, generic webhooks and slack webhooks

pacoguzman:
## Description

- Continuation of the work done at https://github.com/checkr/flagr/pull/240
- Related to issue #214 
- Add a generic webhook to push updates to flags to a user specified endpoint
- Add a slack specific webhook to push updates to a slack channel like this:

54704463-b0502080-4b32-11e9-9c89-2d0b87154200

## Motivation and Context

We needed notifications around who was doing what with the flags we had active, when they were enabled / disabled etc

If this is accepted in a next iteration we would like to generalize the notifiers so we can not use just HTTP based integrations

## How Has This Been Tested?

Manually and with automated tests

## Types of changes

## Checklist:


pkg/config/env.go
```diff
@@ -209,4 +209,21 @@ var Config = struct {
 	// UI path  => localhost:18000/foo"
 	// API path => localhost:18000/foo/api/v1"
 	WebPrefix string `env:"FLAGR_WEB_PREFIX" envDefault:""`
+
+	// NotifyTimeout - for each notifier that uses http.Client to make requests
+	NotifyTimeout time.Duration `env:"FLAGR_NOTIFY_TIMEOUT" envDefault:"30"`
```
zhouzhuojie, line 214: it's duration, so the default value should be something like `30s` - assuming you are referring to 30s of duration.

pkg/config/env.go
```diff
@@ -209,4 +209,21 @@ var Config = struct {
...
+	// NotifyNumAttempts - how many attempts should be made for each notification in case of failure
+	NotifyNumAttempts int `env:"FLAGR_NOTIFY_NUM_ATTEMPTS" envDefault:"10"`
+	// NotifyRetryMin - the minimum amount of time that should pass between retry attempts
+	NotifyRetryMin int `env:"FLAGR_NOTIFY_RETRY_MIN" envDefault:"0"`
```
zhouzhuojie, line 218: if it's `0`, does it mean that no min/max limit? if so, maybe we can add a comment about it.

pkg/config/env.go
```diff
@@ -209,4 +209,21 @@ var Config = struct {
...
+	// NotifyRetryMax - the maximum amount of time that should pass between retry attempts
+	NotifyRetryMax int `env:"FLAGR_NOTIFY_RETRY_MAX" envDefault:"0"`
+
+	// Notify Webhooks
+	NotifyWebhookEnabled bool   `env:"FLAGR_NOTIFY_WEBHOOK_ENABLED" envDefault:"false"`
+	NotifyWebhookURL     string `env:"FLAGR_NOTIFY_WEBHOOK_URL" envDefault:""`
+	// Notify Slack webhooks
```
zhouzhuojie, line 225: For extendibility, I wonder if we need to support multiple channels at this time. The reason is that once we rolled out this change, it's hard to extend the naming of the env to support multiple slack channels. Maybe we should support multiple channels

pkg/handler/crud.go
```diff
@@ -253,6 +255,7 @@ func (c *crud) PutFlag(params flag.PutFlagParams) middleware.Responder {
 	resp.SetPayload(payload)
 
 	entity.SaveFlagSnapshot(getDB(), util.SafeUint(params.FlagID), getSubjectFromRequest(params.HTTPRequest))
+	notify.All(getDB(), util.SafeUint(params.FlagID), notify.UPDATED, notify.FLAG, getSubjectFromRequest(params.HTTPRequest))
```
zhouzhuojie, line 258: A note to myself for the idea of future refactoring - this is a typical hook pattern for both saving the flagSnapshot and notifyAll, maybe good to extract that in the future.

pkg/config/env.go
zhouzhuojie: oh, saw the implementation, so it will use the default settings, maybe worthing adding a comment about the default retry settings.

zhouzhuojie (commented):
still in the review process, will continue sometime later

pkg/config/env.go
pacoguzman: Totally right, I'll change that

pkg/config/env.go
pacoguzman: I'll clarify that with a comment

pkg/config/env.go
pacoguzman: After reading some threads and the current [documentation](https://api.slack.com/messaging/webhooks) it seems to me that the channel now is kind of immutable, so one webhook is bind to a specific channel. So I think we can simplify that removing the channel en var and making the `NotifySlackURL` a slice. Which make me thing we can support multiple webhook urls and not just multiple slack urls, what do you think?

lightsaway:
@pacoguzman @zhouzhuojie  Hello, will this ever be merged?

BYK:
Is there anything we, as the community, do to move this forward? It would be very valuable to have.

pacoguzman:
Sorry but for the time being I won't be able to contribute to this project

BYK:
@pacoguzman I'm happy to take over your awesome work and carry it through the finish line.

@zhouzhuojie is the team willing to accept this feature?

mwarkentin:
Too bad this wasn't added, would be really helpful for us. :(

PR #412, opened 2020-10-20
Add support for kafka auth using SASL with PLAIN mechanism

jaysonsantos:
## Description

This adds variables for authentication and enabling simple ssl without specifying certificates and others

## Motivation and Context

This should enable to use kafka with providers like confluent which uses sasl_plain over ssl.

## How Has This Been Tested?

I've tested it against our cluster which requires authentication with and ssl

## Types of changes

## Checklist:

- [ ] My change requires a change to the documentation. (I am not sure here, does the docs only point to config/env.go?)
- [ ] I have updated the documentation accordingly. (Same as aboce)

### DIFF
--- a/pkg/config/env.go
+++ b/pkg/config/env.go
@@ -134,6 +134,9 @@ var Config = struct {
 	RecorderKafkaKeyFile        string        `env:"FLAGR_RECORDER_KAFKA_KEYFILE" envDefault:""`
 	RecorderKafkaCAFile         string        `env:"FLAGR_RECORDER_KAFKA_CAFILE" envDefault:""`
 	RecorderKafkaVerifySSL      bool          `env:"FLAGR_RECORDER_KAFKA_VERIFYSSL" envDefault:"false"`
+	RecorderKafkaSimpleSSL      bool          `env:"FLAGR_RECORDER_KAFKA_SIMPLE_SSL" envDefault:"false"`
+	RecorderKafkaUsername       string        `env:"FLAGR_RECORDER_KAFKA_USERNAME" envDefault:""`
+	RecorderKafkaPassword       string        `env:"FLAGR_RECORDER_KAFKA_PASSWORD" envDefault:""`
 	RecorderKafkaVerbose        bool          `env:"FLAGR_RECORDER_KAFKA_VERBOSE" envDefault:"true"`
 	RecorderKafkaTopic          string        `env:"FLAGR_RECORDER_KAFKA_TOPIC" envDefault:"flagr-records"`
 	RecorderKafkaRetryMax       int           `env:"FLAGR_RECORDER_KAFKA_RETRYMAX" envDefault:"5"`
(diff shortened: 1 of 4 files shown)


pkg/config/env.go
```diff
@@ -134,6 +134,9 @@ var Config = struct {
 	RecorderKafkaKeyFile        string        `env:"FLAGR_RECORDER_KAFKA_KEYFILE" envDefault:""`
 	RecorderKafkaCAFile         string        `env:"FLAGR_RECORDER_KAFKA_CAFILE" envDefault:""`
 	RecorderKafkaVerifySSL      bool          `env:"FLAGR_RECORDER_KAFKA_VERIFYSSL" envDefault:"false"`
+	RecorderKafkaSimpleSSL      bool          `env:"FLAGR_RECORDER_KAFKA_SIMPLE_SSL" envDefault:"false"`
+	RecorderKafkaUsername       string        `env:"FLAGR_RECORDER_KAFKA_USERNAME" envDefault:""`
+	RecorderKafkaPassword       string        `env:"FLAGR_RECORDER_KAFKA_PASSWORD" envDefault:""`
```
zhouzhuojie, line 139: I think you can add a comment that FLAGR_RECORDER_KAFKA_USERNAME and FLAGR_RECORDER_KAFKA_PASSWORD will only work if FLAGR_RECORDER_KAFKA_SIMPLE_SSL=true

marceloboeira: Should we add something to the docs or the comment is enough?

jaysonsantos: Hey there, according to [confluent docs](https://docs.confluent.io/current/kafka/authentication_sasl/authentication_sasl_plain.html) you can also use SASL with PLAINTEXT which does not require SSL.

jaysonsantos:
I've just rebased the code to force the tests to run as one of them got cancelled.

pkg/handler/data_recorder_kafka.go
```diff
@@ -36,11 +36,19 @@ var NewKafkaRecorder = func() DataRecorder {
...
 	if tlscfg != nil {
 		cfg.Net.TLS.Enable = true
 		cfg.Net.TLS.Config = tlscfg
 	}
+
+	if config.Config.RecorderKafkaUsername != "" && config.Config.RecorderKafkaPassword != "" {
+		cfg.Net.SASL.Enable = true
+	}
+	cfg.Net.SASL.User = config.Config.RecorderKafkaUsername
+	cfg.Net.SASL.Password = config.Config.RecorderKafkaPassword
```
zhouzhuojie, line 50: I think you can also move these two lines into the `if` block, so that if one of them is empty/non-empty, we can minimize the unexpected side effect to the existing kafka config.

pkg/config/env.go
zhouzhuojie: > Should we add something to the docs or the comment is enough?

The doc directly shows the content of this env.go file, so the comment here is good enough.

Also the naming of the env, what do you think of `FLAGR_RECORDER_KAFKA_SASL_USERNAME` and `FLAGR_RECORDER_KAFKA_SASL_PASSWORD`? Just to make it clear that it's using SASL

jaysonsantos:
@zhouzhuojie ready for review again.

zhouzhuojie (approved):

Issue #322, opened 2020-02-17
Add entityId to EntityContext for evaluating constraints


sesquipedalian-dev:
In rolling out a recent feature, I had a case where I wanted to white / black list particular entity IDs from being distributed to the new feature.  It would be handy for Flagr to include the entity ID in the entity context when evaluating constraints, to enable this.  I'm unsure how the property should be named; 'id', 'entityId', 'entity_id', etc. 

## Expected Behavior
You should be able to set a constraint that evaluates the entity ID of the flagr evaluate request.

## Current Behavior
Flagr only includes properties explicitly included in the `entityContext` parameter of eval requests. 

## Possible Solution
Around https://github.com/checkr/flagr/blob/master/pkg/handler/eval.go#L239, before sending the entity context, add an 'id' field containing the value of the entity ID passed in.

zhouzhuojie:
How about `@entityID`? We can even extend this convention to something else that can be derived from the request. 

```
@entityID
@entityType
@timestamp
...
```

Some open questions to consider for records logging:
- Do we include these as entityContext for flagr records logging?
- And since we are already logging these, which means simply putting these into entityContext will have duplicate logging of the evaluation result.

sesquipedalian-dev:
I like the idea of `@<field>`, and including the other fields as well for more flexibility!

PR #418, opened 2020-11-13
Add server-side information to evaluation EntityContext

sesquipedalian-dev:
## Description
This PR sets up options to add additional properties to the `EntityContext` that is used when evaluating which segment on the flag matches the incoming criteria.  Each additional property that is added on the server side is prefixed with '@' to differentiate it from the original client context.  There are two types added: 
- values from the `EvalContext` but aren't available on the `EntityContext`, which might be useful or convenient (e.g. `@entityID` and `@entityType`)
- additional properties configured from the server config / environment - the user may provide a JSON map in the env variable `FLAGR_EVAL_SERVER_ENTITY_CONTEXT` to add these additional properties

## Motivation and Context
https://github.com/checkr/flagr/issues/322

As mentioned in the issue, one case is where you want to pick different variants based on the entity ID provided; in this case the PR change makes it so you don't have to specify the entityID twice in the request.

I added the additional config properties because of the following use case.  We have multiple deployments of flagr (one per 'environment') that use the same flag DB (for configuration management reasons).  Our client requests send up an `env` property in the `EntityContext`, but this means that all our clients have to be configured for that.  If the server specifies the env value, we only have to specify it in the server environment rather than in multiple clients. 

## How Has This Been Tested?
I've only tested it by adding specific go tests so far; I wanted to put up the PR and get feedback.

## Types of changes
- [X] Breaking change (fix or feature that would cause existing functionality to change)
   - potentially if someone is using `EntityContext` properties with an `@` prefix this  change could result in different behavior
   - it also changed the default value of the `FLAGR_RECORDER_KINESIS_AGGREGATE_BATCH_COUNT` config value, because the parser properly validated that the previous default value does not fit into an `int` variable.

## Checklist:

- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/pkg/config/env.go
+++ b/pkg/config/env.go
@@ -42,6 +42,8 @@ var Config = struct {
 	// EvalOnlyMode - will only expose the evaluation related endpoints.
 	// This field will be derived from DBDriver
 	EvalOnlyMode bool `env:"FLAGR_EVAL_ONLY_MODE" envDefault:"false"`
+	// additional values to put in EntityContext when evaluating segments
+	EvalServerEntityContext map[string]string `env:"FLAGR_EVAL_SERVER_ENTITY_CONTEXT" envDefault:"{}"`
 
 	/**
 	DBDriver and DBConnectionStr define how we can write and read flags data.
@@ -151,7 +153,7 @@ var Config = struct {
 	RecorderKinesisFlushInterval       time.Duration `env:"FLAGR_RECORDER_KINESIS_FLUSH_INTERVAL" envDefault:"5s"`
 	RecorderKinesisBatchCount          int           `env:"FLAGR_RECORDER_KINESIS_BATCH_COUNT" envDefault:"500"`
 	RecorderKinesisBatchSize           int           `env:"FLAGR_RECORDER_KINESIS_BATCH_SIZE" envDefault:"0"`
-	RecorderKinesisAggregateBatchCount int           `env:"FLAGR_RECORDER_KINESIS_AGGREGATE_BATCH_COUNT" envDefault:"4294967295"`
+	RecorderKinesisAggregateBatchCount int           `env:"FLAGR_RECORDER_KINESIS_AGGREGATE_BATCH_COUNT" envDefault:"2147483647"`
 	RecorderKinesisAggregateBatchSize  int           `env:"FLAGR_RECORDER_KINESIS_AGGREGATE_BATCH_SIZE" envDefault:"51200"`
 	RecorderKinesisVerbose             bool          `env:"FLAGR_RECORDER_KINESIS_VERBOSE" envDefault:"false"`
 
--- a/pkg/config/env.go
+++ b/pkg/config/env.go
@@ -42,8 +42,22 @@ var Config = struct {
 	// EvalOnlyMode - will only expose the evaluation related endpoints.
 	// This field will be derived from DBDriver
 	EvalOnlyMode bool `env:"FLAGR_EVAL_ONLY_MODE" envDefault:"false"`
-	// additional values to put in EntityContext when evaluating segments
-	EvalServerEntityContext map[string]string `env:"FLAGR_EVAL_SERVER_ENTITY_CONTEXT" envDefault:"{}"`
+	/**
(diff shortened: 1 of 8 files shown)


pkg/entity/fixture.go
```diff
@@ -80,3 +82,47 @@ func GenFixtureSegment() Segment {
 	s.PrepareEvaluation()
 	return s
 }
+
+func GenFixtureSegmentWithAdditionalProperty(propertyName string, entityID string) Segment {
```
zhouzhuojie, line 86: this helper function is too specific that the input parameter doesn't match the name of the function.

how about we reuse `GenFixtureSegment()` in the test, and then just manually add a new constraint the following in the tests? This makes the test looks closer to what exactly it's testing, what do you think?


```
s := GenFixtureSegment()
s.Constraints = append(s.Constraints, 			{
				Model:     gorm.Model{ID: 501},
				SegmentID: 200,
				Property:  propertyName,
				Operator:  models.ConstraintOperatorEQ,
				Value:     fmt.Sprintf(`"%s"`, entityID),
			})

s.PrepareEvaluation()
```

sesquipedalian-dev: Totally agree, implemented this suggestion and got rid of the additional fixture.

pkg/handler/eval_test.go
```diff
@@ -38,6 +40,59 @@ func TestEvalSegment(t *testing.T) {
...
+		}, s)
+
+		assert.NotNil(t, vID)
+		assert.NotEmpty(t, log)
+		assert.False(t, evalNextSegment)
+	})
+
+	t.Run("evalSegment attaches Config.EvalServerEntityContext to EntityContext", func(t *testing.T) {
+		serverEntityContext := map[string]string{"foo": "bar"}
+		config.Config.EvalServerEntityContext = serverEntityContext
```
zhouzhuojie, line 71: please reset the config (ideally with a defer) once the test is done. this will minimize the side effect of future tests

sesquipedalian-dev: good call, done

pkg/handler/eval.go
```diff
@@ -273,6 +273,17 @@ var evalSegment = func(
...
+		m["@entityType"] = evalContext.EntityType
+
+		for propertyName, value := range config.Config.EvalServerEntityContext {
+			if _, ok := m[propertyName]; ok {
+				continue
+			}
+			propertyKey := fmt.Sprintf("@%s", propertyName)
+			m[propertyKey] = value
+		}
+
```
zhouzhuojie, line 286: probably the last ask is to add some documentation for this change, especially what's expected and exactly the property name people can set in their constraints.

e.g. 
```
@entityID
@entityType
@<key> from the env
```

sesquipedalian-dev: Yeah, I'm having a hard time figuring out where to put that documentation.  If you have any ideas?

zhouzhuojie: I think we may need to add docs at the following three places:

1. Overview doc related to the concept of "entity" and add an introduction to "entityContext". https://github.com/checkr/flagr/blame/master/docs/flagr_overview.md#L14
2. Additional description for the entityContext in evalContext https://github.com/checkr/flagr/blob/master/swagger/index.yaml#L444
3. pkg/config/env.go for a little bit more description on how the keys are transformed to `@` prefix in the entityContext

pkg/config/env.go
```diff
@@ -42,6 +42,8 @@ var Config = struct {
 	// EvalOnlyMode - will only expose the evaluation related endpoints.
 	// This field will be derived from DBDriver
 	EvalOnlyMode bool `env:"FLAGR_EVAL_ONLY_MODE" envDefault:"false"`
+	// additional values to put in EntityContext when evaluating segments
```
zhouzhuojie, line 45: the comment can have more information like how the keys can be transitioned into `@` prefix, better if it comes with a simple example.

pkg/handler/eval.go
sesquipedalian-dev: perfect, I'll add something to those sections

lightsaway:
@zhouzhuojie will this be merged anyday ?

Issue #428, opened 2021-01-06
Templates for flag creation


sesquipedalian-dev:
Currently in the flagr API, there is an option to provide a 'template' in the call to create a flag.  The flagr UI uses this with a drop-down menu, where there is one option to create a 'simple boolean flag', that comes pre-populated with a segment and variant.  All of this is hard-coded to a specific example of this simple boolean flag.  I'm wondering if it would be a useful feature to expand the idea of a template. 

I'm imagining from the point of view of the flagr UI, the server could send up a list of templates it supports, which would be displayed in the drop-down menu for creating a flag.  When you picked a different template than the default one, the UI could have additional input boxes for other fields in the template (e.g. rollout percent of the default segment that gets created).  When you submitted the flag creation, it would send the template name and any additional template parameters.  I imagine the template being defined basically like a go template (https://golang.org/pkg/text/template/) for the JSON representation of a flag.   The server would run through the constituent objects of the flag (segments, variants, tags) and save them all in a big GORM transaction, giving you a flag with some default things set up without the user having to do all those steps manually in the UI. 

You'd want the templates to be configurable on the server side so that different users of flagr could set up templates the way they want. This might be a full-fledged resource in the swagger data model, or something loaded from flat files, or maybe just env vars. 

Our team was close to implementing something like this for a validation problem we're looking into (we want to be able to require that flags have a certain type of tag when enabled), but it doesn't quite work.  Curious about gauging if this would be a useful feature to contribute to the project independent of the use case we had.  Here's an experimental branch I was working on: https://github.com/checkr/flagr/tree/sa/flag-template-experiment

## Expected Behavior
People running flagr server should be able to configure additional flag templates with different parameters for creating flags.  When a user selects a different template from the drop-down on flagr UI's home page, they should be prompted for whatever template variables are supported, and when submitted the server would render the template from there. 

## Current Behavior
Currently the 'template' idea is just hard-coded to this simple boolean template, and doesn't allow for configuring different templates for different flagr deployments.  The UI only has a hard-coded field for `description`. 

## Possible Solution

I'm imaging that we would define a simple data structure for templates like so:
```
type FlagTemplateField struct {
  Name string  // identifier for the field
  UIPlaceholderText string // greyed out placeholder text to display in the UI for filling in this field
  Required boolean // if true the user must fill in this field to use the template
}

type FlagTemplate struct {
  Name string // identifier for the template, this would be the value passed for 'template' in the create flag API
[...]

PR #431, opened 2021-01-25
Customizable validation rules for enabled flags

sesquipedalian-dev:
## Description
This changes adds customizable validation rules for _enabled_ flags using https://github.com/Knetic/govaluate.  The rules are specified via new environment variables in the `env.go`; a list of govaluate boolean expressions to be evaluated for flags (a `[]string`), and a connecting operation (either `OR` or `AND`).  When someone uses the API operations `setFlagEnabled` or `putFlag` and the flag is `enabled=true`, we convert the flag into a `map[string]interface{}` so that the properties of the flag are parameters for the govaluate expression.  Then, we iterate through the govaluate expressions in order; if the boolean result of the govaluate expression is `true` then the flag has passed that rule.  Depending on the operation, at least one or all the rules must pass.  If the flag doesn't meet this criteria, a 400 request is returned to the client and the flag update is not saved. 

## Motivation and Context

Our team is trying to keep more aware of how flags are being used on our project.  Specifically we want to associated a flag with an owning project in our JIRA system, so we can provide the teams some info about how the flag is being used (e.g. if the flag in practice always returns the same segment for a period of time).  We landed on using a tag to track this relationship, and we wanted to automate user feedback making sure they assign a tag when the flag is active.  Doing this validation only when a flag is set to enabled seemed like the best way; the user can create and configure the flag without validation up until they enable it for actual use. 

I mentioned this change in this issue at the end: https://github.com/checkr/flagr/issues/428

## How Has This Been Tested?

I initially prototyped using govaluate for this usecase here: https://github.com/sesquipedalian-dev/govaluate-experiment/blob/main/main.go

In addition to adding unit tests covering the new validation rules, I followed a manual test procedure: 
0) environment: Macbook Pro, OSX 10.15.5, chrome browser, go 1.15.6
1) `make run_ui`
2) `FLAGR_ENABLED_FLAG_VALIDATION_RULES='any("regexMatch(Value, \"^JIRA:[a-zA-Z]{3}[a-zA-Z]*$\")", Tags)' make run`
3) in a browser at `localhost:8080` create a new flag with description 'New flag'
4) in the browser open developer tools to follow along with the network calls
5) Click the toggle button to enable the flag, observe you get an error message and the network tab indicates a 400
Screen Shot 2021-01-25 at 10 47 34 AM
6) Reload the page (verify the flag in the DB is not actually enabled) - small UI bug here in that the toggle remains 'enabled' in the UI even though the network call failed
7) Add a tag to the flag matching the critera (e.g. `JIRA:EPLT`)
8) Click the toggle button to enable, observe no error message this time and the network call has a 200 status
9) Remove the tag from the flag
10) change the flag description by adding some characters, and click 'save'.  Observe you again get an error message as in step (5)
11)Add the tag back in and save again, observe no error message

I didn't test other areas of the code too much manually.

## Types of changes

## Checklist:
- [X] My code follows the code style of this project.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ all: deps gen build build_ui run
 rebuild: gen build
 
 test: verifiers
-	@GO111MODULE=on go test -mod=vendor -race -covermode=atomic -coverprofile=coverage.txt github.com/checkr/flagr/pkg/handler
+	@GO111MODULE=on go test -mod=vendor -race -covermode=atomic -coverprofile=coverage.txt github.com/checkr/flagr/pkg/...
 
 .PHONY: benchmark
 benchmark:
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ build_ui:
 run_ui:
 	@cd ./browser/flagr-ui/; npm run serve
 
-run:
+run: build
 	@$(PWD)/flagr --port 18000
 
 start:
--- a/Makefile
+++ b/Makefile
@@ -10,9 +10,8 @@ all: deps gen build build_ui run
 
 rebuild: gen build
 
-# test: verifiers
-test:
-	@GO111MODULE=on go test -mod=vendor -race -covermode=atomic -coverprofile=coverage.txt github.com/checkr/flagr/pkg/...
+test: verifiers
+	@GO111MODULE=on go test -mod=vendor -race -covermode=atomic -coverprofile=coverage.txt github.com/checkr/flagr/pkg/handler
 
 .PHONY: benchmark
 benchmark:
@@ -36,7 +35,7 @@ build_ui:
 run_ui:
 	@cd ./browser/flagr-ui/; npm run serve
 
-run:
+run: build
 	@$(PWD)/flagr --port 18000
 
 start:
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ rebuild: gen build
 
 # test: verifiers
 test:
-	@GO111MODULE=on go test -mod=vendor -race -covermode=atomic -coverprofile=coverage.txt github.com/checkr/flagr/pkg/handler
+	@GO111MODULE=on go test -mod=vendor -race -covermode=atomic -coverprofile=coverage.txt github.com/checkr/flagr/pkg/...
 
 .PHONY: benchmark
 benchmark:
@@ -36,7 +36,7 @@ build_ui:
 run_ui:
 	@cd ./browser/flagr-ui/; npm run serve
 
-run: build
+run:
 	@$(PWD)/flagr --port 18000
 
 start:
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,9 @@ all: deps gen build build_ui run
 
 rebuild: gen build
 
-test: verifiers
-	@GO111MODULE=on go test -mod=vendor -race -covermode=atomic -coverprofile=coverage.txt github.com/checkr/flagr/pkg/...
+# test: verifiers
+test:
+	@GO111MODULE=on go test -mod=vendor -race -covermode=atomic -coverprofile=coverage.txt github.com/checkr/flagr/pkg/handler
 
 .PHONY: benchmark
 benchmark:
--- a/pkg/config/env.go
+++ b/pkg/config/env.go
@@ -217,6 +217,14 @@ var Config = struct {
 	// (putFlag or setFlagEnabled), the flag is evaluated against these
 	// optional rules. The operator indicates if all the rules must pass ('AND' value)
 	// or if at least one rule must pass ('OR' value)
+	//
+	// Example rule:
+	// FLAGR_ENABLED_FLAG_VALIDATION_RULES='any("regexMatch(Value, \"^JIRA:[a-zA-Z]{3}[a-zA-Z]*$\")", Tags)'
+	// this requires that the flag have at least one tag that matches a regex starting with 'JIRA:' and ending with a
+	// project name consisting of a-z characters, like 'EPLT'
+	//
+	// to use multiple rules, separate each rule with ':::'.
+	//
 	EnabledFlagValidationRules     []string `env:"FLAGR_ENABLED_FLAG_VALIDATION_RULES" envSeparator:":::"`
 	EnabledFlagValidationOperation string   `env:"FLAGR_ENABLED_FLAG_VALIDATION_OPERATOR" envDefault:"AND"`
 }{}
--- a/pkg/config/env.go
+++ b/pkg/config/env.go
@@ -212,4 +212,14 @@ var Config = struct {
 	// UI path  => localhost:18000/foo"
 	// API path => localhost:18000/foo/api/v1"
 	WebPrefix string `env:"FLAGR_WEB_PREFIX" envDefault:""`
+
+	// rules for validating flags on enable - when a flag is updated or enabled
+	// (putFlag or setFlagEnabled), the flag is evaluated against these
+	// optional rules. The operator indicates if all the rules must pass ('AND' value)
+	// or if at least one rule must pass ('OR' value)
+	EnabledFlagValidationRules     []string `env:"FLAGR_ENABLED_FLAG_VALIDATION_RULES" envSeparator:","`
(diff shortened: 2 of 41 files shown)


Makefile
```diff
@@ -11,7 +11,7 @@ all: deps gen build build_ui run
 rebuild: gen build
 
 test: verifiers
-	@GO111MODULE=on go test -mod=vendor -race -covermode=atomic -coverprofile=coverage.txt github.com/checkr/flagr/pkg/...
+	@GO111MODULE=on go test -mod=vendor -race -covermode=atomic -coverprofile=coverage.txt github.com/checkr/flagr/pkg/handler
```
zhouzhuojie, line 14: we still need to test basically all the packages under `pkd/...`

sesquipedalian-dev: whoops I thought I reverted that

pkg/config/env.go
```diff
@@ -212,4 +212,22 @@ var Config = struct {
...
 	// API path => localhost:18000/foo/api/v1"
 	WebPrefix string `env:"FLAGR_WEB_PREFIX" envDefault:""`
+
+	// rules for validating flags on enable - when a flag is updated or enabled
+	// (putFlag or setFlagEnabled), the flag is evaluated against these
+	// optional rules. The operator indicates if all the rules must pass ('AND' value)
+	// or if at least one rule must pass ('OR' value)
+	//
+	// Example rule:
+	// FLAGR_ENABLED_FLAG_VALIDATION_RULES='any("regexMatch(Value, \"^JIRA:[a-zA-Z]{3}[a-zA-Z]*$\")", Tags)'
```
zhouzhuojie, line 222: if you already have `any` or `all`, is the `OR`/`AND` operator still needed?

pkg/config/env.go
```diff
@@ -212,4 +212,22 @@ var Config = struct {
 	// UI path  => localhost:18000/foo"
 	// API path => localhost:18000/foo/api/v1"
 	WebPrefix string `env:"FLAGR_WEB_PREFIX" envDefault:""`
+
+	// rules for validating flags on enable - when a flag is updated or enabled
```
zhouzhuojie, line 216: I think `Enabled...` is not special here. Maybe we can just call it `CustomFlagValidationRules`, and uses a hook pattern to call this `CustomFlagValidationRules` whenever there's a write path change to the specified flag.

pkg/config/env.go
```diff
@@ -212,4 +212,22 @@ var Config = struct {
...
+	// (putFlag or setFlagEnabled), the flag is evaluated against these
+	// optional rules. The operator indicates if all the rules must pass ('AND' value)
+	// or if at least one rule must pass ('OR' value)
+	//
+	// Example rule:
```
zhouzhuojie, line 221: If we introduce generic validation rules, I worried that we may need to carry on these validation rules' compatibility forever, and it's error-prone given that there're four layers of syntax for the Rules:

1. regex
2. flag's properties as the goevaluate context
3. custom goevaluate syntax like `any` / `all`
4. EnabledFlagValidationOperation for multiple rules

There's a steep learning curve without a place to try-and-test these rules - especially in production.

How about we start from the dedicated validation first? For example, we can have, for example

- `CustomTagsValidationRules`
- `CustomFlagKeyValidationRules`
- ...

pkg/config/env.go
sesquipedalian-dev: In the spike we reviewed to decide on this approach, the team thought that having multiple rules + an `OR/AND` would make writing any individual rule less complicated.  Of course, since govaluate supports those operations you could just stuff the entire thing into the evaluation expression.  I'd be inclined to go that route (take out OR/AND operator) since it takes out a layer of complexity (you have to understand only govaluate).

pkg/config/env.go
sesquipedalian-dev: I think I agree.  What we specifically want to allow is users to create a fresh flag (especially since the UI only supports a default / 'simple boolean' new flag template (https://github.com/checkr/flagr/issues/428)), then tweak it until it's ready to turn on, and only when it's turned on require the additional validation step.  I think that's consistent with renaming things as you suggested.

pkg/config/env.go
sesquipedalian-dev: > There's a steep learning curve without a place to try-and-test these rules
To me, using the test framework seemed like a relatively good way to try out things: https://github.com/checkr/flagr/pull/431/files#diff-a5d5bc2d830216d981ead1c1e20ca3ab19c22874b17bb775fc3188d2992ded58R96

> How about we start from the dedicated validation first?
I think starting with `CustomTagsValidationRules` would handle the use case I have.  I'm a little unsure how it would handle if I had two different formats of tags that I wanted to enforce, and that seems like a common enough use case after you have one tag format.  Are the semantics of `CustomTagsValidationRules` that any tag would have to meet at least one of the validation rules? Because that's not what I want; if the flag has no tags it would pass the rules.  Are the semantics that a flag must have one tag that meets the rules? Then I can't enforce two different types of tags.  I guess I'd lean towards removing `4. EnabledFlagValidationOperation for multiple rules` and making `EnabledFlagValidationRules` a single string. 

> custom goevaluate syntax like any / all
The custom govaluate functions seem like something that should be in the govaluate library rather than us adding them here, but I could understand if they don't want to add in everything to the shared lib.  Still I should probably offer a PR.

pkg/handler/enabled_flag_validation.go
```diff
@@ -0,0 +1,201 @@
...
+	createEvaluationRule.Do(func() {
+		for _, structExpressionString := range config.Config.EnabledFlagValidationRules {
+			functions := map[string]govaluate.ExpressionFunction{
+				"regexMatch": RegexMatchFunc,
+				"any":        Any,
+			}
+
+			expression, err := govaluate.NewEvaluableExpressionWithFunctions(structExpressionString, functions)
+			if expression == nil || err != nil {
+				evaluationRulesSetupError = err
```
MJBonifacio, line 137: Looks like this will exit early if we encounter any error without fully building our evaluation rule set? Should we log an error of some kind or at least `return nil, err` to propagate whatever failed up?

pkg/handler/enabled_flag_validation.go
```diff
@@ -0,0 +1,201 @@
...
+			return false, fmt.Errorf("Unknown evaluation rule result type (%v)", result)
+		}
+
+		if config.Config.EnabledFlagValidationOperation == config.FlagValidationOperationOR {
+			valid = valid || resultBool
+			if valid {
+				break
+			}
+		} else if config.Config.EnabledFlagValidationOperation == config.FlagValidationOperationAND {
+			valid = valid && resultBool
```
MJBonifacio, line 176: Perhaps not too consequential, but I think we can early exit if any validation operation fails.

pkg/handler/enabled_flag_validation.go
```diff
@@ -0,0 +1,201 @@
...
+		if err != nil {
+			return flag.NewPutFlagDefault(500).WithPayload(ErrorMessage("%s", err))
+		}
+		if !valid {
+			opMsg := "all"
+			if config.Config.EnabledFlagValidationOperation == config.FlagValidationOperationOR {
+				opMsg = "one"
+			}
+
+			return flag.NewPutFlagDefault(400).WithPayload(ErrorMessage("Flag failed %s of these validation rules: %s", opMsg, config.Config.EnabledFlagValidationRules))
```
MJBonifacio, line 197: Is this necessarily true? Only a single condition has to fail in a series of "And" evaluations to fail the validation.

pkg/handler/crud.go
```diff
@@ -237,6 +237,10 @@ func (c *crud) PutFlag(params flag.PutFlagParams) middleware.Responder {
 		f.Notes = *params.Body.Notes
 	}
 
+	if err := validateFlagIfEnabled(f); err != nil {
```
MJBonifacio, line 240: Is it possible there is `enabled` is being set to true in params? like so https://github.com/checkr/flagr/blob/cff1f323fd82b12ab2792ce4b2c3fd49f3e30cce/swagger_gen/models/put_flag_request.go#L30

PR #433, opened 2021-02-16
Add dropdown by state to filter flags

juanantoniocid:
## Description

In the home page we currently have a first row to add new flags and a second row to search for flags. We have modified the second row in order to show a dropdown list in which we can see the values \[`All the flags`, `Enabled flags`, `Disabled flags`\] being the value `All the flags` the one selected by default.

As the [current API](https://checkr.github.io/flagr/api_docs/#operation/findFlags) supports the boolean parameter `enabled` to filter flags by enabled and disabled, we only needed to change the GUI to add the filter.

## Motivation and Context

The number of flags is growing and as this is a service shared by all of our customers, it is expected it will continue growing in the future. To do it easier to work with Flagr we want to provide a way of filtering flags by status (enabled / disabled).

## How Has This Been Tested?

1. I added four flags: `Flag Y1`, `Flag Y2`, `Flag Z1` and `Flag Z2`.
2. I enabled flags `Flag Y1` and `Flag Z1`.
3. I selected `Enabled flags` and only `Flag Y1` and `Flag Z1` were shown (as expected).
4. I selected `Disabled flags` and only `Flag Y2` and `Flag Z2` were shown (as expected).
5. I selected `All the flags` and the four flags were shown (as expected).
6. I filtered by the term "Z" and only `Flag Z1` and `Flag Z2` were shown (as expected).
7. I selected `Enabled flags` and only `Flag Z1` was shown (as expected).
8. I selected `Disabled flags` and only `Flag Z2` was shown (as expected).
9. I removed the term "Z"only `Flag Z1` and `Flag Z2` were shown (as expected).

## Types of changes

## Checklist:

### DIFF
--- a/browser/flagr-ui/src/components/Flags.vue
+++ b/browser/flagr-ui/src/components/Flags.vue
@@ -37,12 +37,26 @@
           </el-row>
 
           <el-row>
-            <el-input
-              placeholder="Search a flag"
-              prefix-icon="el-icon-search"
-              v-model="searchTerm"
-              v-focus
-            ></el-input>
+            <el-col :span="19">
+              <el-input
+                placeholder="Search a flag"
+                prefix-icon="el-icon-search"
+                v-model="searchTerm"
+                v-focus
+              ></el-input>
+            </el-col>
+            <el-col :span="4" :offset="1">
+              <template>
+                <el-select v-model="statusValue" placeholder="Filter by state">
+                  <el-option
+                    v-for="item in statusFilter"
+                    :key="item.value"
+                    :label="item.label"
+                    :value="item.value">
+                  </el-option>
+                </el-select>
+              </template>
+            </el-col>
           </el-row>
 
           <el-table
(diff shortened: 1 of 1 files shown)


zhouzhuojie:
Thanks for the PR. What do you think if we just add a filter for the "Enabled" column in the flags table? 

E.g.
https://element.eleme.cn/#/en-US/component/table#filter

For example from element's documentation:
image


I think we may just need to add the filter and something like "All", "Enabled", and "Disabled" for the `filters`.

```
      :filters=...
      :filter-method=...
      filter-placement=...
```

juanantoniocid:
> Thanks for the PR. What do you think if we just add a filter for the "Enabled" column in the flags table?
> 

For sure! I'll take a look and I'll be back with the suggested changes.

juanantoniocid:
I have done the commit [`44b6b7e`](https://github.com/checkr/flagr/pull/433/commits/44b6b7eb0875517f92da9048bdc535612a04825a) to filter flags by state, using a filter for the "Enabled" column in the flags table as suggested above.

zhouzhuojie:
lgtm

zhouzhuojie (approved):

PR #564, opened 2024-01-01
Change to use pure golang sqlite driver for better portability

nothing0012:
## Description
Update the sqlite dep to be golang only, this helps the portability and reduce the dependency on cgo

Tests and Benchmark of the new sqlite3 driver:

image

## Types of changes

## Checklist:


pkg/entity/fixture.go
```diff
@@ -3,7 +3,6 @@ package entity
 import (
 	"github.com/openflagr/flagr/swagger_gen/models"
 
-	_ "gorm.io/driver/sqlite" // sqlite driver
```
nothing0012, line 6: this was already imported and used in db.go

go.mod
```diff
@@ -56,13 +55,12 @@ require (
 
 require (
 	cloud.google.com/go/pubsub v1.30.0
-	github.com/goccy/go-json v0.10.2
```
nothing0012, line 59: go-json was not used.

running `make vendor` to tidy this

PR #597, opened 2024-12-16
V1.1.17

harkirat321:
## Types of changes

## Checklist:


Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN make build_ui
 ######################################
 # Prepare go_builder
 ######################################
-FROM golang:1.21-alpine as go_builder
+FROM golang:1.23.4-alpine as go_builder
```
noodle-rubrik, line 14: Does something from 1.21-alpine in go_builder end up in the final image and get scanned? I'd expect that we have to update the alpine used in the final image instead on line 21.

(unless, that ends up using the same `alpine` as go_builder, since it doesn't specify...???)

harkirat321: Go dependencies are resolved and build from the alpine in line 14 as the `Run make build` is executed on line 19. 
Line 21 alpine is just for fulfilling runtime dependencies and is very light weight.

harkirat321: Also, `FROM alpine` would anyways pull the latest stable image of alpine. So i believe it is not mandatory to mention the version there?

PR #621, opened 2025-06-11
POC: CheckList

serh11d:
## Description
**Backend**
Introduced a new operator _CHECKLIST_, it expects JSON array of objects
```JSON
[{
 "c": 1,
 "v": "value",
 "d": "desc"
}]
```
_c_ - checked, used 0, 1 to reduce payload
_v_ - string value
_d_ - description, if not present then it has 0 value to reduce payload

The idea is to filter such array of objects with `c == 1` and map values to array with overriding operator to _IN_.

I wasn't able to determine is value only stored in text format and cached in memory as object instance or it's parsed every time a flag is requested. Also I wasn't able to get any stable performance results on my laptop they fuzzy from time to time, guess this should be done on a dedicated device, possibly I will do that on one of my old (it's even better as sensitive for performance) SBCs. So format is discussible, it could be plain text like `1value,desc\n` (first checked, then value until comma[, desc] and \n as element separator) or even base64 encoded binary linked list.

**UI**
Added new operator type _CHECKLIST_
Added new (...) buttons on the right of _value_ inputs.
image

Added edit collection dialog that supports _IN_, _NOTIN_, _CHECKLIST_ operators.
image
image

Implemented basic validation.
Additional sugar should not be difficult like all check/uncheck, open _IN_, _NOTIN_ in _CHECKLIST_ mode.

I also want to add that I don’t have significant knowledge of either Golang or Vue, and changes in the code may differ greatly from your expectations :stuck_out_tongue_winking_eye:

## Motivation and Context
Very often you have to work with lists of IDs like GUID or hash that cannot be identified simply by reading. Sometimes there are dozens of them. Sometimes it is necessary to briefly exclude certain elements from the list. All of the above together introduces certain inconveniences when working directly with the JSON representation of the list (array of elements).

## How Has This Been Tested?
**Backend** Unit test + local env.
**UI** Local environment

## Types of changes

## Checklist:


zhouzhuojie:
I like the direction, it's basically a sugar for an array of OR constraints. 

One thing we need to be clear about the UX, is that for example, in your case, how does the `CHECKLIST` match the context?

1. What's the name of the property in the context to match the checklist?
2. Assuming we have the property, what's the logic and order of the checklist? If it matches the deny, does it always deny? If it matches allow in the checklist, does it always allow?
3. I feel more like a UX/UI change and use the primitive constraints that's already defined in the backend, wdyt?

image

sergiyd:
@zhouzhuojie If I correctly understood the question, _CHECKLIST_ is only for _OR_ where checked/unchecked stays included this value into result _IN_ constraint or not. To have an opposite effect, there should be implementation of _NOTCHECKLIST_ that will be transformed into _NOTIN_ in back-end. I started with _IN_ because it is more in demand according to my observations.

PR #686, opened 2026-05-19
perf: short-circuit EvalCache reload when no new flag_snapshot

zhouzhuojie:
## Problem

`EvalCache.reloadMapCache()` runs every 3 seconds by default and unconditionally performs a **full eager-load** of all flags (with segments, variants, constraints, distributions, and tags) from the database — even when zero data has changed. At scale, this creates unnecessary database load, CPU/memory churn from deserialization and constraint parsing, and GC pressure.

## Solution

Short-circuit the full reload by checking whether any new `flag_snapshot` rows have been created since the last load. Every CRUD mutation that affects evaluation data creates a flag snapshot, so `SELECT MAX(id) FROM flag_snapshots` is a reliable change indicator.

### Coverage verification

Every mutation handler that affects the eval cache calls `entity.SaveFlagSnapshot(...)`:

| Handler | Operation | Snapshot created? |
|---|---|---|
| `CreateFlag` / `PutFlag` / `SetFlagEnabledState` | Flag CRUD | ✅ |
| `DeleteFlag` / `RestoreFlag` | Flag delete/restore | ✅ |
| `CreateTag` / `DeleteTag` | Tag association | ✅ |
| `CreateSegment` / `PutSegment` / `PutSegmentsReorder` | Segment CRUD | ✅ |
| `DeleteSegment` | Segment soft delete | ✅ |
| `CreateConstraint` / `PutConstraint` / `DeleteConstraint` | Constraint CRUD | ✅ |
| `PutDistributions` | Distribution batch update | ✅ |
| `CreateVariant` / `PutVariant` / `DeleteVariant` | Variant CRUD | ✅ |

18 mutation handlers total — all create a snapshot. Tag association changes (hard deletes from the `flags_tags` join table) are also covered since `DeleteTag` calls `SaveFlagSnapshot`.

### What about soft deletes?

Soft-deletes set `deleted_at` without bumping `updated_at` on the parent entity, but `SaveFlagSnapshot` is always called after the delete, so `MAX(id)` on `flag_snapshots` still increments.

### What about future model changes?

If someone adds a new entity that affects evaluation data, the corresponding handler should call `SaveFlagSnapshot` (which is already the established convention for the existing 18 handlers). If it does not, the cache will not detect the change — same as if they bypass `SaveFlagSnapshot` for any existing handler. The startup self-check (querying DB metadata for all table names) is not part of this PR but could be added separately.

## Changes

- **`pkg/handler/eval_cache.go`** — Added `lastSnapshotMaxID uint` field to `EvalCache`. In `reloadMapCache()`, run a lightweight `SELECT COALESCE(MAX(id), 0) FROM flag_snapshots` via GORM before the full fetch. Short-circuit when it has not changed.
- **`pkg/handler/eval_cache_test.go`** — Added `TestReloadMapCacheShortCircuit` that verifies the fetch is skipped on repeated calls and re-enabled after a new snapshot. Existing tests reset `lastSnapshotMaxID` to ensure they always load.
- **`pkg/handler/export_test.go`** — Same reset for the singleton.
- **`docs/flagr_overview.md`** — Documented the EvalCache and its short-circuit behavior in the Architecture section.

## Performance

Before: full multi-join eager load of every flag every 3s.
After: one primary-key index lookup (`MAX(id)`) per tick. At ~1ms per query, this is negligible.

### DIFF
--- a/docs/flagr_overview.md
+++ b/docs/flagr_overview.md
@@ -45,8 +45,8 @@ The definitions of the following concepts are in [API doc](https://openflagr.git
 
 There are three components in the flagr, Flagr Evaluator, Flagr Manager, and Flagr Metrics.
 
-- Flagr Evaluator. Flagr evaluator evaluates the incoming requests.
-- Flagr Manager. Flagr manager is the CRUD gateway. All the mutations of flags happen here.
-- Flagr Metrics. Flagr metrics is the data pipeline to collect evaluation results. Currently Flagr only supports Kafka as the pipeline.
+- **Flagr Evaluator**. Flagr evaluator evaluates the incoming requests. It maintains an in-memory cache (`EvalCache`) of all flags, segments, variants, constraints, distributions, and tags — pre-parsed and ready for fast evaluation. The cache is refreshed periodically (default every 3s, configurable via `FLAGR_EVALCACHE_REFRESHINTERVAL`). To avoid unnecessary database load, the reload short-circuits when no new `flag_snapshot` rows have been created, since every mutation handler that affects evaluation data also creates a snapshot.
+- **Flagr Manager**. Flagr manager is the CRUD gateway. All the mutations of flags happen here.
+- **Flagr Metrics**. Flagr metrics is the data pipeline to collect evaluation results. Currently Flagr only supports Kafka as the pipeline.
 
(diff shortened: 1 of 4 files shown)


zhouzhuojie:
will close https://github.com/openflagr/flagr/issues/685

PR #742, opened 2026-07-09
docs: accurate behavioral contracts and leaner guide set

zhouzhuojie:
## Description

Tighten and correct user-facing docs so integrator rules match the evaluator, with a clearer home page and less meta baggage.

### Accuracy (canonical: `flagr_behavioral_contracts.md`)

- **Segment evaluation:** first segment whose **constraints** match stops evaluation. Low rollout can leave `variantKey` empty; it does **not** fall through to later segments.
- **Blank vs stream:** empty `variantKey` is not the same as “no stream row.” Missing / disabled / no-segments skip recording; no-match and rollout-empty can still emit evaluation rows when gates are open.
- **Eval volume vs exposure:** eval volume is fine for rollouts, kill switches, server-side config, and many dashboards. **Rigid A/B denominators** (people who actually saw the surface) should use explicit exposure after render.
- **entityID:** omit → random id injected **before** bucketing (non-sticky). Stickiness requires a stable client-sent id.

### Ownership and cleanup

- Renamed **`contracts.md` → `flagr_behavioral_contracts.md`** so the filename matches the title. No redirect stub.
- Removed **`link-compatibility.md`** (hand-maintained route catalog) and the CONTRIBUTING pointer to it. Prefer delete-on-rename and update in-repo links.
- Other guides link to behavioral contracts instead of restating mini-essays.
- Restored nested indent in use-case example templates.
- README architecture links to overview for bucketing (no inline CRC32 dump).
- Eval-only: JSON drivers as the usual path; full export path; `snapshots/max_id` breadcrumb restored.

## Motivation and Context

Overview and related pages taught an incorrect rollout fallthrough model and conflated blank assignment with “no streaming record.” Contracts wording was also too absolute about never counting eval volume. This PR makes behavioral contracts the single source for cross-cutting rules, softens the A/B denominator guidance, and drops docs that only existed to catalog redirects.

## How Has This Been Tested?

- Docs-only (no Go/UI runtime code).
- Cross-checked against `pkg/handler/eval.go` (`evalSegment` stops after constraint match; early `BlankResult` skips `logEvalResult`) and `pkg/entity/distribution.go`.
- Grepped for residual fallthrough / blank=no-stream claims and stale `contracts.md` / `link-compatibility` references.

## Types of changes

## Checklist:


zhouzhuojie:
Addressed the pending review comments in a0d4582e:

1. **`flagr_use_cases.md` indent** (3 templates) — restored nested list hierarchy under Variants / Segment / Distribution (and attachment nesting under variants).
2. **`contracts.md` filename** — renamed to **`flagr_behavioral_contracts.md`** so the path matches the “Behavioral contracts” title. Sidebar, README, and content links updated. Kept a short **stub** at `contracts.md` so old `#/contracts` bookmarks still resolve; new links use `#/flagr_behavioral_contracts`.

PR #761, opened 2026-07-26
feat(mcp): Add built-in MCP server for AI agent flag management

marceloboeira:
## Description

AI coding assistants are becoming a primary interface for developers. Adding a built-in MCP server lets agents manage flags directly without leaving the editor or hand-crafting API calls.

The server is embedded in the Flagr binary at /mcp on the same port as the REST API, using Streamable HTTP transport. Single binary, single port, no sidecars.

Disabled by default. Set FLAGR_MCP_ENABLED=true to opt in. When off the route simply doesn't exist, so no extra attack surface.

27 tools covering the full flag lifecycle: CRUD for flags, segments, variants, and constraints, plus distributions, tags, and evaluation. Built with the official Go MCP SDK. Integration tests run across the full docker-compose matrix with a disabled-mode test verifying the route is absent when the flag is off.

## How Has This Been Tested?

Besides all the unit/integration tests, I tested the e2e functionality by running opencode locally and pointing to the built MCP (see images).

### Input
Screenshot 2026-07-26 at 16 11 47

### Result

Screenshot 2026-07-26 at 16 12 03

### Created Flag

Screenshot 2026-07-26 at 16 12 35

Screenshot 2026-07-26 at 16 12 19

## Types of changes

## Checklist:


marceloboeira:
@zhouzhuojie do you want to take a look? in theory there is nothing breaking as it's disabled by default.

pkg/mcp/tool_flag.go
```diff
@@ -0,0 +1,517 @@
...
+		Enabled:            in.Enabled,
+		DataRecordsEnabled: in.DataRecordsEnabled,
+		EntityType:         in.EntityType,
+		Notes:              in.Notes,
+		CreatedBy:          "mcp",
+		UpdatedBy:          "mcp",
+	}
+
+	db := getDB()
+	if err := db.Create(f).Error; err != nil {
```
zhouzhuojie, line 121: **[bug]** Nearly all MCP write tools mutate flags/segments/variants/constraints/distributions/tags via raw GORM without `commitFlagMutation` / `WriteFlagSnapshotTx` (only `duplicate_flag` writes a snapshot). EvalCache intentionally short-circuits reloads when `MAX(flag_snapshots.id)` is unchanged (`pkg/handler/eval_cache.go` documents that every API mutation that affects eval data creates a snapshot). After MCP mutations, both `evaluate_flag` and the HTTP evaluation API can keep serving stale in-memory flag data until some unrelated REST mutation creates a new snapshot.

**Suggestion:** Route MCP mutations through the same `commitFlagMutation` path (or always call `entity.SaveFlagSnapshot` / `WriteFlagSnapshotTx` after successful writes) so snapshot IDs advance and EvalCache refreshes. Prefer sharing handler/entity helpers rather than duplicating GORM logic.

swagger_gen/restapi/configure_flagr.go
```diff
@@ -45,7 +46,23 @@ func configureAPI(api *operations.FlagrAPI) http.Handler {
...
+	baseHandler := setupGlobalMiddleware(api.Serve(setupMiddlewares))
+
+	// NOTE: This is a hand-edit in generated code. If you run `make swagger`,
+	// re-apply this block: import "github.com/openflagr/flagr/pkg/mcp" and
+	// wrap baseHandler with an http.ServeMux that routes /mcp to the MCP handler.
+	// Mount MCP server on /mcp when enabled.
+	if mcp.MCPEnabled() {
+		mcpSrv := mcp.New()
+		mux := http.NewServeMux()
+		mux.Handle("/mcp", mcpSrv.Handler())
```
zhouzhuojie, line 59: **[bug]** `/mcp` is mounted on a top-level `ServeMux` that wraps the already-middleware-wrapped API handler, so the MCP handler does not run through `setupGlobalMiddleware`. That bypasses JWT/Basic auth, panic recovery, request logging, metrics, CORS, and `WebPrefix` stripping. Docs admit the auth bypass; if someone enables MCP on a production host that relies on JWT/Basic for admin APIs, unauthenticated clients can create/delete flags and change rollouts. Docs also incorrectly claim elsewhere that “auth/logging is shared with the HTTP API.”

**Suggestion:** Mount MCP inside (or after) the same global middleware chain, or add explicit MCP auth (shared JWT/Basic and/or a dedicated token). At minimum, apply recovery + logging middleware to `/mcp`, honor `WebPrefix`, and make the security model a first-class warning (default off is not enough for accidental enablement).

pkg/mcp/tool_constraint.go
```diff
@@ -0,0 +1,276 @@
...
+	FlagID        uint `json:"flag_id"`
+	SegmentID     uint `json:"segment_id"`
+	Distributions []struct {
+		VariantID  uint   `json:"variant_id"`
+		VariantKey string `json:"variant_key"`
+		Percent    uint   `json:"percent"`
+	} `json:"distributions"`
+}
+
+func (s *Server) handleSetDistributions(ctx context.Context, req *mcp.CallToolRequest) (*mcp.CallToolResult, error) {
```
zhouzhuojie, line 229: **[bug]** `set_distributions` docs/schema say percents must sum to 100, but the handler never validates sum, segment ownership (`flag_id`+`segment_id`), or that each `variant_id`/`variant_key` belongs to the flag. REST `PutDistributions` does all of this via `validatePutDistributions` and ownership checks. Agents can write invalid distribution rows that break evaluation until fixed.

**Suggestion:** Reuse `validatePutDistributions` / ownership helpers (or equivalent checks): require sum==100, verify segment under flag, verify each variant under flag and key match, reject bad input before writing.

pkg/mcp/tool_constraint.go
```diff
@@ -0,0 +1,276 @@
...
+	var in setDistributionsInput
+	if err := decodeArgs(req.Params.Arguments, &in); err != nil {
+		return errResult("invalid arguments: %v", err), nil
+	}
+
+	db := getDB()
+
+	// Delete existing distributions.
+	if err := db.Where("segment_id = ?", in.SegmentID).Delete(&entity.Distribution{}).Error; err != nil {
```
zhouzhuojie, line 238: **[bug]** `set_distributions` deletes existing distributions then creates new ones without a transaction. If a later `Create` fails, the segment is left with partial or empty distributions—worse than the pre-call state. REST wraps delete+create in `commitFlagMutation`’s transaction.

**Suggestion:** Perform clear+insert in a single DB transaction (ideally via the shared mutation helper that also snapshots).

pkg/mcp/tool_segment.go
```diff
@@ -0,0 +1,183 @@
...
+	db := getDB()
+	seg := &entity.Segment{}
+	if err := entity.PreloadConstraintsDistribution(db).First(seg, in.SegmentID).Error; err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return errResult("segment %d not found", in.SegmentID), nil
+		}
+		return errResult("failed to get segment: %v", err), nil
+	}
+
+	seg.RolloutPercent = in.RolloutPercent
```
zhouzhuojie, line 124: **[bug]** `update_segment` schema marks `rollout_percent` optional, but the handler always assigns `seg.RolloutPercent = in.RolloutPercent`. A call that only updates `description` (or omits rollout) silently sets rollout to 0, which can disable traffic for that segment.

**Suggestion:** Use a pointer (`*uint`) or presence check for `rollout_percent` and only update when provided; same pattern REST uses for optional fields where applicable.

pkg/mcp/tool_variant.go
```diff
@@ -0,0 +1,167 @@
...
+		v.Key = util.SafeString(in.Key)
+	}
+	if in.Attachment != nil {
+		v.Attachment = entity.Attachment(in.Attachment)
+	}
+
+	if err := v.Validate(); err != nil {
+		return errResult("%v", err), nil
+	}
+	if err := db.Save(v).Error; err != nil {
```
zhouzhuojie, line 136: **[bug]** After updating a variant key, REST `PutVariant` calls `validatePutVariantForDistributions` to sync `distributions.variant_key`. MCP `update_variant` only saves the variant, leaving distribution rows with the old key. Evaluation and distribution listings can disagree with the variant table.

**Suggestion:** After a successful key change, update related distributions’ `variant_key` in the same transaction (reuse `validatePutVariantForDistributions`).

pkg/mcp/tool_variant.go
```diff
@@ -0,0 +1,167 @@
...
+	var in struct {
+		FlagID    uint `json:"flag_id"`
+		VariantID uint `json:"variant_id"`
+	}
+	if err := decodeArgs(req.Params.Arguments, &in); err != nil {
+		return errResult("invalid arguments: %v", err), nil
+	}
+
+	db := getDB()
+	if err := db.Delete(&entity.Variant{}, in.VariantID).Error; err != nil {
```
zhouzhuojie, line 153: **[bug]** `delete_variant` deletes by ID with no check for non-zero distributions. REST `DeleteVariant` runs `validateDeleteVariant` and refuses deletion when a distribution still has non-zero percent (and cleans zero-percent rows). MCP can orphan/inconsistently remove variants that are actively used in rollouts.

**Suggestion:** Call the same delete validation logic as the REST handler before deleting.

pkg/mcp/tool_flag.go
```diff
@@ -0,0 +1,517 @@
...
+
+	tx := db.Begin()
+	defer func() {
+		if r := recover(); r != nil {
+			tx.Rollback()
+		}
+	}()
+
+	newFlag := &entity.Flag{
+		Key:                src.Key + "_copy",
```
zhouzhuojie, line 354: **[bug]** `duplicate_flag` hardcodes `Key: src.Key + "_copy"` instead of generating a unique key via `entity.CreateFlagKey` / API-style optional key input. Duplicating the same source twice collides on the unique key index; long keys can exceed the 64-char limit; unlike REST, there is no way to supply a target key/description, and `Enabled` is forced false while REST preserves source enabled state (docs claim disabled-on-duplicate, so that part may be intentional, but key handling is still broken).

**Suggestion:** Align with `pkg/handler/crud_duplicate.go`: accept optional key/description, run `CreateFlagKey`, use `ApplyFlagTemplate` / `commitFlagMutation`, and return a clear 400 on key conflicts.

pkg/mcp/tool_segment.go
```diff
@@ -0,0 +1,183 @@
...
+	var in struct {
+		FlagID    uint `json:"flag_id"`
+		SegmentID uint `json:"segment_id"`
+	}
+	if err := decodeArgs(req.Params.Arguments, &in); err != nil {
+		return errResult("invalid arguments: %v", err), nil
+	}
+
+	db := getDB()
+	if err := db.Delete(&entity.Segment{}, in.SegmentID).Error; err != nil {
```
zhouzhuojie, line 146: **[bug]** Multiple tools require `flag_id` in the schema but ignore it for authorization/ownership: e.g. `delete_segment` deletes solely by `segment_id`; constraint create/update/delete and distribution tools similarly trust IDs alone. REST validates segment/variant/constraint ownership under the given flag. Wrong `flag_id` with a valid foreign ID still mutates another flag’s resources, which is confusing for agents and weakens any future multi-tenant controls.

**Suggestion:** For every nested mutation, verify the resource belongs to `flag_id` (reuse `validateSegmentOwnership` / `validateVariantOwnership` / `validateConstraintOwnership`).

docs/ai-mcp.md
```diff
@@ -0,0 +1,389 @@
...
+**What the agent does (tool calls):**
+
+1. `create_flag` → `{ "key": "new-checkout", "description": "New checkout flow" }`
+2. `create_variant` → `{ "flag_id": 1, "key": "control" }`
+3. `create_variant` → `{ "flag_id": 1, "key": "treatment", "attachment": { "ui_version": "v2" } }`
+4. `create_segment` → `{ "flag_id": 1, "rollout_percent": 50, "description": "US users" }`
+5. `create_constraint` → `{ "flag_id": 1, "segment_id": 1, "property": "country", "operator": "EQ", "value": "\"US\"" }`
+6. `set_distributions` → `{ "flag_id": 1, "segment_id": 1, "distributions": [{ "variant_id": 1, "variant_key": "control", "percent": 50 }, { "variant_id": 2, "variant_key": "treatment", "percent": 50 }] }`
+
+### Setting up a gradual rollout with constraints
```
zhouzhuojie, line 244: **[bug]** Cline integration documents spawning `./flagr` as an MCP **command** (stdio) server. `Server.Run` implements stdio transport but nothing in `cmd/` or configure path ever starts it—`FLAGR_MCP_ENABLED` only mounts Streamable HTTP at `/mcp`. That Cline config starts a normal HTTP Flagr process speaking the wrong protocol on stdio, so the documented integration does not work.

**Suggestion:** Either wire a real stdio entrypoint (e.g. `flagr mcp` / env mode that only runs `mcp.Server.Run`) or remove/replace the Cline command-mode example with the remote HTTP transport.

pkg/config/env.go
```diff
@@ -45,6 +45,10 @@ var Config = struct {
 	// UIEnabled controls whether the Flagr UI is served.
 	// Set to false for backend-only deployments where the UI is not needed.
 	UIEnabled bool `env:"FLAGR_UI_ENABLED" envDefault:"true"`
+	// MCPEnabled enables the built-in MCP server (stdio transport).
```
zhouzhuojie, line 48: **[suggestion]** `MCPEnabled` comment still describes “stdio transport” and “stdin/stdout,” but production path is Streamable HTTP at `/mcp`. Misleading for operators reading `env.go` / generated env docs.

**Suggestion:** Update the comment (and any env guide text) to describe HTTP `/mcp`, off-by-default, and auth caveats.

pkg/mcp/server.go
```diff
@@ -0,0 +1,125 @@
...
+			nil,
+		),
+	}
+	s.registerTools()
+	return s
+}
+
+// Run starts the MCP server over stdio. Blocks until ctx is cancelled or
+// stdin is closed. Call as a goroutine.
+func (s *Server) Run(ctx context.Context) error {
```
zhouzhuojie, line 43: **[suggestion]** `Run` (stdio), `StartEvalCache`, and related helpers look unfinished/dead: EvalCache is already started from `handler.setupEvaluation`, and stdio `Run` is never called. Dead API surface increases confusion about supported transports.

**Suggestion:** Remove unused exports until stdio is productized, or document and wire a single supported transport clearly.

swagger_gen/restapi/configure_flagr.go
```diff
@@ -45,7 +46,23 @@ func configureAPI(api *operations.FlagrAPI) http.Handler {
 	api.ServerShutdown = config.ServerShutdown
 
 	handler.Setup(api)
-	return setupGlobalMiddleware(api.Serve(setupMiddlewares))
+
+	baseHandler := setupGlobalMiddleware(api.Serve(setupMiddlewares))
+
+	// NOTE: This is a hand-edit in generated code. If you run `make swagger`,
```
zhouzhuojie, line 52: **[suggestion]** Hand-editing `configure_flagr.go` is the go-swagger extension point (“safe to edit”), but AGENTS.md generally says never hand-edit `swagger_gen/`. The NOTE about re-applying after `make swagger` is easy to miss and will break MCP on regen if someone regenerates without the block.

**Suggestion:** Keep the hand-edit (acceptable for this file) but add a CI check that `/mcp` registration remains when MCP is enabled, and/or extract mounting to a tiny non-generated helper invoked from configure to minimize merge/regen risk. Document the exception next to the swagger workflow in AGENTS.md.

pkg/mcp/tool_flag.go
```diff
@@ -0,0 +1,517 @@
...
+type createFlagInput struct {
+	Key                string `json:"key"`
+	Description        string `json:"description"`
+	Enabled            bool   `json:"enabled"`
+	DataRecordsEnabled bool   `json:"data_records_enabled"`
+	EntityType         string `json:"entity_type"`
+	Notes              string `json:"notes"`
+}
+
+func (s *Server) handleCreateFlag(ctx context.Context, req *mcp.CallToolRequest) (*mcp.CallToolResult, error) {
```
zhouzhuojie, line 98: **[suggestion]** Tool handlers reimplement domain rules instead of calling existing CRUD/handler functions. Besides snapshot/validation gaps already filed, this will drift (entity types on create, notifications, delete soft-delete side effects, etc.). Long-term maintenance cost is high for 27 tools.

**Suggestion:** Thin MCP tools should map args → existing handler/entity operations (or a shared service layer used by both REST and MCP), returning MCP-shaped errors.

opencode.jsonc
```diff
@@ -0,0 +1,11 @@
+{
```
zhouzhuojie, line 1: **[suggestion]** Root `opencode.jsonc` commits a local IDE MCP client config pointing at `localhost:18000`. This is developer-machine tooling, not product code, and will confuse contributors who do not run Flagr on that port or use OpenCode.

**Suggestion:** Drop from the PR or put under an examples/ path / docs snippet instead of repo root config.

docs/ai-mcp.md
```diff
@@ -0,0 +1,389 @@
+---
```
zhouzhuojie, line 1: **[suggestion]** New page is linked from README but not added to VitePress sidebar (`docs/.vitepress/config.mts`), so it is harder to discover in the docs site. Env var is also only documented in this page, not integrated with the usual `flagr_env` narrative beyond the struct embed.

**Suggestion:** Add an “AI & MCP” nav item and cross-link from self-host/auth docs with the security caveats.

pkg/mcp/tool_flag.go
```diff
@@ -0,0 +1,517 @@
...
+	}
+
+	db := getDB().Unscoped()
+	flags := []entity.Flag{}
+
+	if in.Enabled != nil {
+		db = db.Where("enabled = ?", *in.Enabled)
+	}
+	if in.DescriptionLike != "" {
+		db = db.Where("lower(description) like ?", fmt.Sprintf("%%%s%%", in.DescriptionLike))
```
zhouzhuojie, line 175: **[nit]** `list_flags` interpolates `description_like` into a SQL `LIKE` pattern with `fmt.Sprintf("%%%s%%", ...)`. Values are bound as parameters (not injection), but `%`/`_` in user input act as LIKE wildcards without escaping—same class of footgun as many list filters.

**Suggestion:** Escape LIKE metacharacters if parity with careful REST filtering is desired; otherwise document behavior.

swagger_gen/restapi/configure_flagr.go
```diff
@@ -45,7 +46,23 @@ func configureAPI(api *operations.FlagrAPI) http.Handler {
...
+	if mcp.MCPEnabled() {
+		mcpSrv := mcp.New()
+		mux := http.NewServeMux()
+		mux.Handle("/mcp", mcpSrv.Handler())
+		mux.Handle("/", baseHandler)
+		logrus.Info("mcp server enabled at /mcp (streamable HTTP)")
+		return mux
+	}
+
+	return baseHandler
```
zhouzhuojie, line 65: this code better moved outside of the swagger_gen code, as it's more for auto generated code, you can find a better place for this like in the handler package.

pkg/mcp/server.go
```diff
@@ -0,0 +1,125 @@
...
+			&mcp.Implementation{
+				Name:    "flagr-mcp",
+				Version: "0.1.0",
+			},
+			nil,
+		),
+	}
+	s.registerTools()
+	return s
+}
```
zhouzhuojie, line 39: this server opens up the AUTH protection, we might need to carefully design or control the who and how can they access the /mcp

zhouzhuojie (commented):
Given the auth complexity, I'm wondering if it's just simpler to create a SKILL.md and ask the agent to check the swagger api and ask the user to provide AUTH so that the agent can do its job without the need of this mcp endpoint

PR #764, opened 2026-07-30
feat: read-only UI for eval-only mode (json_file/json_http)

fukuanru:
## Description

Makes the UI usable on eval-only deployments (`FLAGR_DB_DBDRIVER=json_file` / `json_http`) as a **read-only flag browser**. Today those deployments still serve the UI static assets (`FLAGR_UI_ENABLED` defaults to true), but every CRUD call returns 501, so the UI is effectively broken.

> **Reworked after review** (see [review comment](https://github.com/openflagr/flagr/pull/764#issuecomment-5288708849)): the first cut registered a second EvalCache-backed `CRUD` implementation. That dual implementation is gone — the UI is now a **client of the existing `GET /export/eval_cache/json`** endpoint, with one mapper at the frontend API boundary.

**Backend (small)**
- Eval-only registered surface is unchanged from `main`: health + evaluation + export. No CRUD routes.
- Write denial: a single `evalOnlyDeny` middleware (`pkg/config/middleware.go`) returns **403** with a message pointing at the JSON source for any non-GET request under `/api/v1/flags` (WebPrefix-aware). Every CRUD write endpoint lives under that prefix, so one check covers all 19 write ops — previously they returned a misleading 501.
- `GET /health` gains `evalOnlyMode: boolean` (swagger source + regenerated `swagger_gen/`) for UI mode discovery.
- `GET /flags/snapshots/max_id` is untouched — no fingerprint overload; the endpoint keeps its documented monotonic-snapshot-ID semantics.

**Frontend**
- New `src/api/evalCache.ts`: fetches the export dump (GitOps `entity.Flag` shape, PascalCase) and maps it once to the swagger camelCase `Flag` the UI already speaks. The export JSON shape is not changed. Module-level cache of the mapped dump (list → detail → back is one fetch; concurrent readers share one in-flight request). Flags are sorted by id (the export iterates a Go map, so order is random per fetch); segment order is kept as exported because in json mode source order **is** evaluation order.
- `crud.ts` read functions branch on `evalOnlyMode`: the list refetches the export on mount (no server-side change token), `getFlag`/`listAllTags` read the mapped dump (tags deduped by value), entity types / snapshots / deleted flags resolve `[]` locally. Components and pages are unchanged.
- The server mode is resolved **before first paint**: `main.ts` awaits `/health` (bounded by a 1.5s timeout, fail-open) before `app.mount`, so a read-only deployment never flashes editable controls.
- Read-only banner; Create Flag / Deleted Flags / Flag Management / History tab hidden; inputs disabled via a `readonly` prop threaded through the section components; `?tab=history` deep links land on Config.
- **Debug Console stays available** — evaluation works in eval-only mode and is the main reason to open the UI on an eval edge node.
- UI hiding is UX only; the backend 403 is the enforcement.

No new env var: the UI reads the same export endpoint that already exists, so there is no new exposure surface, and auth middlewares apply unchanged.

Design doc: `docs/plans/2026-07-30-001-feat-eval-only-readonly-ui-plan.md` (rework noted in-doc).

## Motivation and Context

The GitOps flow (#699) made `json_file` / `json_http` the recommended eval-edge shape, but operators lost all UI visibility into what those nodes are actually serving. This PR restores browsing and live evaluation debugging on those nodes without reintroducing any write path — the JSON source stays the single source of truth.

## How Has This Been Tested?

- `make test` — golangci-lint, swagger validate, `go test ./pkg/...` all green. New `TestEvalOnlyDenyMiddleware` covers: 403 for all write shapes under `/flags` including un-normalized paths (`//api/v1/flags`, dot segments), pass-through for reads/evaluation/export/OPTIONS, WebPrefix variants (unset, `/flagr`, `/flagr/`, `/`, multi-segment), and no-op when eval-only is off.
- `make build-ui` — ESLint, vue-tsc, 91 vitest tests, Vite production build all green. New tests: `evalCache` mapper against a PascalCase fixture mirroring `pkg/handler/testdata/sample_eval_cache.json`, dump cache/in-flight dedupe/404/tag dedupe, and the eval-only branches in `crud.ts`; the existing `initServerMode` and history deep-link guard tests remain.
- New Playwright spec `e2e/readonly.spec.ts` (5 tests): banner visible + write affordances hidden on the list, detail page browsable-but-disabled with Debug Console present, history deep link lands on Config, and two late-`/health` tests asserting the UI recovers through the export path when the mode resolves after the 1.5s mount bound. Un-stubbed `/api/v1` calls are aborted so regressions fail loudly.
- `make build-docs` — VitePress build green.
- Manually ran the built binary with a `json_file` source and verified with curl:
  - `GET /api/v1/health` → `{"evalOnlyMode":true,"status":"OK"}`
  - `POST /api/v1/flags`, `PUT /api/v1/flags/1/enabled` → `403` `"Flagr is running in read-only (eval-only) mode: ..."`
  - `GET /api/v1/flags` → `501` (unregistered, as on `main`; the UI never calls it in this mode)
  - `GET /api/v1/export/eval_cache/json` → dump served
  - `POST /api/v1/evaluation` → normal evaluation result

### Screenshots (json_file source)

Flags list — read-only banner, no Create Flag, no Deleted Flags:

read-only flags list

Flag page — inputs disabled, no save/delete/reorder/add controls, no History tab, Debug Console available:

read-only flag detail

`?tab=history` deep link lands on Config instead of an empty pane:

history deep link

## Types of changes

## Checklist:


marceloboeira (approved):

zhouzhuojie:
I kind of prefer improving the export endpoint and create some mapper, essentially what UI needs is just a dump of the exported json to sit on readonly mode. I asked AI to draft some detailed plan according to the context of this PR and my understanding of your intent. response below:




##########################################

##########################################

-------------------


## Request changes: serve the read-only UI from the existing export dump

The product goal is right — eval-only (`json_file` / `json_http`) already serves the UI assets, but every CRUD call 501s, so operators cannot browse what the node is actually evaluating or use the Debug Console. Keeping writes rejected, History hidden, and Debug Console available is the correct UX.

The backend shape is heavier than that goal. `readOnlyCRUD` is a second 32-method `CRUD` implementation: a parallel `FindFlags` query engine (the UI list is `GET /flags` with **no** query params; search is client-side), sub-resource GETs the UI never calls (`…/segments`, `…/variants`, `…/constraints`, `…/distributions`, `GET /flags/{id}/tags`), 19 write stubs that exist only to satisfy the interface, and a content fingerprint stuffed into `GET /flags/snapshots/max_id` so `listFlagsIfStale` does not have to change.

That is a permanent dual-implementation tax. Every future CRUD method now has two homes, and filter semantics will drift from GORM.

Eval-only already has a fully populated data plane: **`GET /api/v1/export/eval_cache/json`**. That *is* the live EvalCache. The UI in this mode only needs to read that dump. Please drop the second CRUD and load the read-only UI from export + a mapper instead.

### Why not “just swap the URL”

Export is the GitOps / `entity.Flag` JSON (PascalCase, `{ "Flags": [ … ] }`, GORM fields). The Vue app speaks swagger camelCase (`id`, `key`, `rolloutPercent`, `variantID`). Pointing `get('/flags')` at export would leave every field `undefined`.

That is a **mapper**, not a new API. Do not change the export JSON shape to help the UI — that file is the GitOps source. Map once at the API boundary and keep the rest of the tree on `Flag`.

### Plan

**Keep from this PR**

- `evalOnlyMode` on `GET /health` (mode discovery; one justified swagger change).
- `initServerMode()` before first paint, fail-open, 1.5s bound.
- Read-only banner; hide Create / Deleted Flags / Flag Management / History; `readonly` props on the section components; Debug Console stays.
- History deep-link guard (`applyDeepLink` + the `Flag.vue` watcher).
- Docs: eval-only is a read-only browser; JSON source remains the only write path.

**Drop**

- `pkg/handler/crud_readonly.go` and its tests.
- `GetAllFlags` / `ContentFingerprint` / `flagsFingerprint` (they exist to feed the twin CRUD and to fake `max_id`).
- Re-registering the full CRUD surface in eval-only `Setup`.
- Overloading `GET /flags/snapshots/max_id` as a hash. Swagger says snapshot max id; external pollers may assume monotonic. The list cache does not need it (below).

**Backend leftover (small)**

Eval-only `Setup` stays health + evaluation + export, as today.

Writes: I still like **403** with a pointer at the JSON source better than a bare 501. That does **not** need a `CRUD` impl — one shared deny helper (or middleware on mutating `/api/v1/flags*`) is enough. No `FindFlags` clone.

**Frontend: one adapter, then the existing types**

New file, e.g. `browser/flagr-ui/src/api/evalCache.ts` (or a thin branch inside `crud.ts` — one module either way):

1. `GET /export/eval_cache/json` → `{ Flags: ExportedFlag[] }`.
2. `mapExportedFlag(e): Flag` (and nested segment / variant / tag / constraint / distribution mappers). Ignore GORM-only fields (`DeletedAt`, `SnapshotID`, `FlagID`, `SegmentID`, timestamps we do not display).
3. Return the existing swagger-shaped `Flag`. `normalizeFlag` stays unchanged.

Suggested field map (from the sample export / JSON spec):

| Export (`entity.Flag`) | UI (`Flag`) |
|---|---|
| `ID` | `id` |
| `Key` | `key` |
| `Description` | `description` |
| `Enabled` | `enabled` |
| `Notes` | `notes` |
| `CreatedBy` / `UpdatedBy` / `UpdatedAt` | `createdBy
| `DataRecordsEnabled` | `dataRecordsEnabled` |
| `EntityType` | `entityType` |
| `Tags[].ID` / `Value` | `tags[].id` / `value` |
| `Variants[].ID` / `Key` / `Attachment` | `variants[]
| `Segments[].ID` / `Description` / `Rank` / `RolloutP
| `Constraints[].Property` / `Operator` / `Value` | sa
| `Distributions[].Percent` / `VariantID` / `VariantKe

Do **not** sprinkle `flag.Key ?? flag.key` through components.

**How the existing UI reads get satisfied** (no new CRUD routes)

The read-only UI needs a fully nested flag tree. Export is already that.

| Current call | Eval-only |
|---|---|
| `listFlags()` / `listFlagsIfStale()` | Fetch export,
| `getFlag(id)` | Find in the in-memory dump (reuse /
| `listAllTags()` | Unique `value`s from the same dump
| `listEntityTypes()` | `[]` (already unused in this m
| `listFlagSnapshots()` | Never called if History stay

`loadFlagPageContext` should take one dump and derive `{ flag, allTags, entityTypes: [] }` instead of three CRUD GETs.

Keep a module-level cache of the mapped dump so list → detail → back does not refetch three times. Invalidate on list remount (or always refetch export there). Do not invent a server-side change token.

**Writes / enforcement**

UI hiding stays UX only. Backend 403 (or 501 if you skip the deny helper) is the gate. No change to eval, auth, or export filters.

**Tests**

- Mapper unit tests against a fixture that looks like `pkg/handler/testdata/sample_eval_cache.json` (PascalCase, nested segments, GORM nulls).
- `initServerMode` tests you already have.
- History deep-link guard you already have.
- Adapter: eval-only list/detail/tags come from export; writable mode still uses CRUD.
- Please drop the `filterFlags` / sub-resource / fingerprint Go tests that only exist for `readOnlyCRUD`.
- A single Playwright path would help more than another filter test: banner visible, Create hidden, Debug Console still evaluates, `POST /flags` is 403.

**Docs**

Eval-only registered surface stays: health (`evalOnlyMode`), evaluation, `GET /export/eval_cache/json`. The UI is a **client of export** in that mode, not a reason to grow the CRUD surface. Update the plan / behavioral-contracts wording that currently promises read-only CRUD GETs.

### What I am not asking for

- Do not teach every Vue component the GitOps JSON schema.
- Do not add `?format=api` to export or camelCase the dump.
- Do not implement a “thin” `FindFlags`+`GetFlag` CRUD twin as a compromise unless you hit a real blocker in the mapper — export already is the fully exported JSON the UI needs to read.

Happy to re-review on that shape. The UI chrome in this PR can stay almost as-is.

fukuanru:
@zhouzhuojie Thanks for the detailed review — agreed on all counts, and the rework is pushed in ee57c24. The UI is now a client of `GET /export/eval_cache/json` with one mapper at the API boundary; the second CRUD implementation is gone.

**Dropped, as requested**
- `pkg/handler/crud_readonly.go` + tests (−652 lines). Eval-only `Setup` is back to health + evaluation + export, exactly as on `main`.
- `GetAllFlags` / `GetByFlagID` / `ContentFingerprint` / `flagsFingerprint` — `eval_cache.go` and `eval_cache_fetcher.go` are byte-identical to `main` again.
- The `max_id` fingerprint overload. The endpoint keeps its documented monotonic semantics; the list page simply refetches the export on mount (no invented change token).

**Kept**
- `evalOnlyMode` on `GET /health`, `initServerMode()` before first paint (1.5s bound, fail-open), banner / hidden write affordances / `readonly` props, history deep-link guard, Debug Console.
- The 403 for writes — now a single `evalOnlyDeny` middleware in `pkg/config/middleware.go` on non-GET `/api/v1/flags*` (WebPrefix-aware), no `CRUD` impl involved. Every write op lives under that prefix, so one check covers all 19.

**The mapper** — `browser/flagr-ui/src/api/evalCache.ts`, per your field map. The export JSON shape is untouched, no `flag.Key ?? flag.key` anywhere, and `normalizeFlag` is unchanged. `crud.ts` read functions branch on `evalOnlyMode`, so components/pages didn't change at all. A module-level cache of the mapped dump makes list → detail → back one fetch (concurrent readers share one in-flight request), and `loadFlagPageContext` derives `{ flag, allTags, entityTypes: [] }` from that single dump.

Two implementation details worth flagging:
- The export handler iterates the `idCache` map, so flag order is random per fetch — the adapter sorts by ID before the usual newest-first reverse.
- Segment order is deliberately **not** sorted by rank: in json mode the source order is the evaluation order, so the UI shows segments exactly as the node evaluates them.

**Tests** — the `filterFlags`/sub-resource/fingerprint Go tests went away with the impl. New: `TestEvalOnlyDenyMiddleware` (403 shapes, pass-throughs, WebPrefix, off-mode no-op), mapper/adapter vitest suites against a PascalCase fixture mirroring `sample_eval_cache.json`, and the Playwright path you suggested — `e2e/readonly.spec.ts`: banner visible, Create hidden, detail browsable-but-disabled with Debug Console present, history deep link lands on Config. The backend 403 is asserted in the Go middleware test (the shared e2e backend runs writable mode). Also smoke-tested the built binary with a `json_file` source: health reports the mode, writes 403, export + evaluation work.

Docs (`flagr_behavioral_contracts.md` and friends) now say the eval-only surface stays health/evaluation/export and the UI is a client of export in that mode; the plan doc records the rework. PR description updated to match. Ready for another look!


---

**Follow-up (14da1e4b):** after the rework I ran three rounds of adversarial verification against this branch and hardened what they surfaced — all confirmed by tests before pushing:

- **Late `/health` race.** `main.ts` mounts fail-open after a 1.5s bound; if `/health` resolved later than that on a real eval-only server, the first fetch went down the (unregistered) CRUD path and the list/detail pages spun forever with no retry. The pages now watch the `evalOnlyMode` flip and refetch through the export path, and a failed list load shows an explicit "couldn't load" empty state with a Retry button instead of a misleading "no flags yet". Two new Playwright tests reproduce the race (health delayed past the mount bound, CRUD stubs returning 501) and assert the recovery — verified they fail on the pre-fix commit.
- **`evalOnlyDeny` path normalization.** The 403 originally matched the raw request path, so `POST //api/v1/flags` or `/api/v1/x/../flags` fell through to the router's 501. No write was ever possible (the CRUD handlers aren't registered in eval-only mode), but the documented enforcement should hold on its own. The deny now checks both normalization orders — clean-then-strip-prefix, and strip-then-clean to mirror `http.StripPrefix` + the swagger router — which also covers dot segments spanning a multi-segment `FLAGR_WEB_PREFIX` and trailing-slash prefixes. Table-driven tests cover WebPrefix unset, `/flagr`, `/flagr/`, `/`, `/a/b`, `/a/b/c`, and `/flagr//`.
- **E2e spec honesty.** `readonly.spec.ts` now registers a catch-all `route.abort()` for `/api/v1/**` before the specific stubs, so any request the read-only UI shouldn't make fails the test loudly instead of being silently served by the writable e2e backend.

All suites green after the follow-up: `make test`, ESLint/vue-tsc/91 vitest, 5 Playwright tests, plus real-binary curl checks in `json_file` mode (bypass shapes 403, evaluation/health unaffected, normal sqlite mode unchanged).

fukuanru:
@zhouzhuojie Please help to review again. We need this read only UI for our gitops :)

zhouzhuojie (approved):

PR #771, opened 2026-08-20
refactor: simplify eval-only UI and reject prefix-escape paths

zhouzhuojie:
## Summary

Follow-up to #764. Tighten the eval-only read-only UI and the write-deny path without changing the product contract (read-only UI, 403 on flag writes, export as the data plane).

**UI**
- `evalOnlyMode` is the single source of truth. `crud.ts` derives HTTP vs export reads from it (no second `FlagReadSource` type).
- Rename `api/evalCache.ts` → `api/evalOnly.ts` (eval-only read plane, not a generic EvalCache client).
- Keep the 1.5s `Promise.race` before first paint and the list/detail `evalOnlyMode` watchers. A late `/health` still refetches through the export dump; history deep links snap back to Config.

**Middleware**
- `rejectDotDotPath` runs before JWT/basic and the flags deny: any `..` in the path is **401** (same status the JWT tests already used for GET `/..` and `/api/v1/health/..`). Stops a prefix-escape from skipping a whitelist and then being Clean()'d into a write.
- `evalOnlyDeny` is a struct (`flagsPath`, `denyMsg`) installed inside `StripPrefix`, matching clean `/api/v1/flags` with `HasSafePrefix`.

**Docs** — behavioral contracts, env, eval-only plan as-built.

## Test plan

- [x] GitHub Actions on this PR: `unit_test`, `ui_lint`, `docs_build`, `e2e_test`, `integration_test` all green
- [x] Go: `TestEvalOnlyDenyMiddleware`, `TestRejectDotDotPath`, `TestIsFlagsAPIPath`, JWT 401 paths including `/..` and `/api/v1/health/..`
- [x] Vitest: `crud.test.ts`, `evalOnly.test.ts`, `serverMode.test.ts`, `flagPage.test.ts`

### E2E coverage (`browser/flagr-ui/e2e/readonly.spec.ts`)

CI `e2e_test` passed. The spec is **unchanged in behavior** vs #764 (helper moved only). It still covers:

- banner + hidden writes on the list
- detail dump data, disabled inputs, no History, Debug Console
- `?tab=history` → Config
- late `/health` recovery on list and detail (the watchers)

That is the right layer for **UI** changes. It is **not** coverage for the new middleware: the spec stubs `/health` and `/export/eval_cache/json` and aborts other `/api/v1/**`, so it never hits `rejectDotDotPath` or `evalOnlyDeny`. Those stay in `pkg/config` unit tests (as the spec already notes for 403s). No extra Playwright case is needed for `..` path attacks.


marceloboeira (approved):

zhouzhuojie:
@fukuanru some more refactoring, should be fully compatible with your previous change

